PR #22, opened 2023-06-05
Added NET_ADMIN capability to evpn bridge container

udhay-104:
NET_ADMIN capability is needed to avoid permission errors while executing read-write netlink commands. Adding this in opi-evpn-bridge since it shares the same network as leaf1.


glimchb (approved):
lgtm
thanks for adding more description

PR #23, opened 2023-06-06
create: use {resource}_id field if provided by user

glimchb:


glimchb:
@udhay-104 FYI

PR #30, opened 2023-06-15
Adding CreateVpc and DeleteVpc for VRF creation and deletion.

udhay-104:
Used linux "ip link" command as a place holder. This will be replaced by proper GO NETLINK package.
Need to work on moving the static configuration in docker-compose file to a simple grpc client which calls the created functions to configure the nodes. This will be done once the minimum required functions are created.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  contents: read
+
 jobs:
   call:
     uses: opiproject/opi-smbios-bridge/.github/workflows/commitlint.yml@main


pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+// CreateVpc executes the creation of the VRF/VPC
+func (s *Server) CreateVpc(_ context.Context, in *pb.CreateVpcRequest) (*pb.Vpc, error) {
+	log.Println("CreateVpc: Received request to create new VPC/VRF from client")
+	log.Printf("Received Info: %v\n", in)
+	// check required fields
+	if err := fieldbehavior.ValidateRequiredFields(in); err != nil {
+		log.Printf("error: %v", err)
+		return nil, err
+	}
+	// Check whether VPC/VRF exists
```
glimchb, line 42: missing block here dealing with {resource}_id, see below:
```
	// see https://google.aip.dev/133#user-specified-ids
	resourceID := resourceid.NewSystemGenerated()
	if in.SubnetId != "" {
		err := resourceid.ValidateUserSettable(in.SubnetId)
		if err != nil {
			log.Printf("error: %v", err)
			return nil, err
		}
		log.Printf("client provided the ID of a resource %v, ignoring the name field %v", in.SubnetId, in.Subnet.Name)
		resourceID = in.SubnetId
	}
	in.Subnet.Name = fmt.Sprintf("//network.opiproject.org/vpc/%s", resourceID)
```

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+	vpc, ok := s.Vpcs[in.Vpc.Name]
+	if ok {
+		log.Printf("VPC/VRF %v already exists\n", in.Vpc.Name)
+		return vpc, nil
+	} else {
```
glimchb, line 47: else not needed, you have return above

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+		args := strings.Split(command, " ")
+		cmd := exec.Command(args[0], args[1:]...)
+		b, err := cmd.CombinedOutput()
+		if err != nil {
+			log.Printf("Command Failed: %v", err)
+			return in.Vpc, err
+		}
+		fmt.Printf("%s\n", b)
+	}
+	response := proto.Clone(in.Vpc).(*pb.Vpc)
```
glimchb, line 64: missing `response.status = ...`

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+	vpc, ok := s.Vpcs[in.Vpc.Name]
+	if ok {
+		log.Printf("VPC/VRF %v already exists\n", in.Vpc.Name)
+		return vpc, nil
+	} else {
+		log.Printf("Creating new VPC/VRF %v\n", in.VpcId)
+	}
+	// Temporary code to configure linux VRF via exec commands. To be replaced by netlink package
+	var commands []string
+	commands = append(commands, fmt.Sprintf("sudo ip link add %v type vrf table %v", in.VpcId, in.Vpc.Spec.V4RouteTableNameRef))
```
glimchb, line 52: as you stated, better to replace this with netlink package

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+
+// DeleteVpc executes the deletion of the VRF/VPC
+func (s *Server) DeleteVpc(_ context.Context, in *pb.DeleteVpcRequest) (*emptypb.Empty, error) {
+	log.Println("DeleteVpc: Received request to delete a VPC/VRF from client")
+	log.Printf("Received Info: %v\n", in)
+	if err := fieldbehavior.ValidateRequiredFields(in); err != nil {
+		log.Printf("error: %v", err)
+		return nil, err
+	}
+	// Check whether VPC/VRF exists
```
glimchb, line 79: missing block here to validate full name, see below:
```
	// Validate that a resource name conforms to the restrictions outlined in AIP-122.
	if err := resourcename.Validate(in.Name); err != nil {
		log.Printf("error: %v", err)
		return nil, err
	}
```

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+	vpc, ok := s.Vpcs[in.Name]
+	if !ok {
+		err := status.Errorf(codes.NotFound, "VRF/VPC %s doesn't exist", in.Name)
```
glimchb, line 82: missing allow missing block here:
```
	if !ok {
		if in.AllowMissing {
			return &emptypb.Empty{}, nil
		}
		err := status.Errorf(codes.NotFound, "unable to find key %s", in.Name)
		log.Printf("error: %v", err)
		return nil, err
	}
```

pkg/evpn/evpn.go
```diff
@@ -25,6 +27,76 @@ type Server struct {
...
+		log.Printf("Error: %v\n", err)
+		return nil, err
+	}
+	log.Printf("Deleting VPC/VRF %v\n", in.Name)
+	log.Println(vpc)
+	command := fmt.Sprintf("sudo ip link del %v", vpc.Name)
```
glimchb, line 88: as you know, let's try use netlink package

glimchb (changes requested):
few comments inside
plus I would like to see the docker compose changes to test this is actually working or not

PR #37, opened 2023-06-23
ci: remove static bridge config and use grpc

glimchb:


glimchb:
used to calculate mac:
```
python -c "import base64; print(base64.b64encode('\xaa\xbb\xcc\x00\x00\x01'))"
qrvMAAAB
```

PR #48, opened 2023-07-06
Implement tunnel RPCs

glimchb:
Need for vxlan
Plugged into Bridge

Create/Get/Delete/Update functions first implementation

Few todos and gaps inside still...


glimchb:
example of object creation:
```
	testTunnel = pb.Tunnel{
		Spec: &pb.TunnelSpec{
			LocalIp: &pc.IPAddress{
				Af:     pc.IpAf_IP_AF_INET,
				V4OrV6: &pc.IPAddress_V4Addr{V4Addr: 336860161},
			},
			Encap: &pc.Encap{
				Type: pc.EncapType_ENCAP_TYPE_VXLAN,
				Value: &pc.EncapVal{
					Val: &pc.EncapVal_Vnid{Vnid: 100},
				},
			},
		},
	}
```

venkatmahalingam (approved):

mardim91 (commented):
I see that the tunnels are pnly atached to a bridge. I guess this only for L2VNIs right ? 

Even in that case the L2VPNs will not work because the bridge needs to be attached to a VRF even if that VRF is not bind to a L3VNI. This is the way that FRR understand L2VPNs os a non L3VNI VRF is needed also in the L2VPN case

I will need to double check that because I am not 100% sure

glimchb:
> bridge needs to be attached to a VRF

bridge is attached already to VRF, which is VPC object... what is missing ?

mardim91:
Are we talking here about implementing L3VPNs symmetric IRB or L2VPNs ?

PR #97, opened 2023-08-08
fix(port): raise new error

### LINKED ISSUE #95: Port: for `Access` type, the `LogicalBridge` list must have only one item. If not then we need to throw error

glimchb:
Fixes #95


pkg/evpn/port.go
```diff
@@ -51,6 +51,13 @@ func (s *Server) CreateBridgePort(_ context.Context, in *pb.CreateBridgePortRequ
 		log.Printf("Already existing BridgePort with id %v", in.BridgePort.Name)
 		return obj, nil
 	}
+	// for Access type, the LogicalBridge list must have only one item
```
mardim91, line 54: ok seems right to me

PR #99, opened 2023-08-08
refactor(svi): move from cloud APIs to evpn-gw APIs

glimchb:


pkg/evpn/svi.go
```diff
@@ -27,88 +27,122 @@ import (
...
+			fmt.Printf("Failed to set MAC on link: %v", err)
 			return nil, err
 		}
-		if err := netlink.LinkSetMaster(vxlan, brdev); err != nil {
-			fmt.Printf("Failed to add vxlan to bridge: %v", err)
+	}
+	// Example: ip address add <svi-ip-with prefixlength> dev <link_svi>
+	for _, gwip := range in.Svi.Spec.GwIpPrefix {
+		fmt.Printf("Assign the GW IP address %v to the SVI interface %v", gwip, vlandev)
+		myip := make(net.IP, 4)
```
venkatmahalingam, line 99: do we need to create an issue to make the code generic enough to handle ipv4 & ipv6 addresses?

pkg/evpn/svi.go
```diff
@@ -130,51 +164,51 @@ func (s *Server) DeleteTunnel(_ context.Context, in *pb.DeleteTunnelRequest) (*e
...
+	// use netlink to find vlan
+	vlan, err := netlink.LinkByName(resourceID)
 	if err != nil {
 		err := status.Errorf(codes.NotFound, "unable to find key %s", resourceID)
 		log.Printf("error: %v", err)
 		return nil, err
 	}
 	// bring link down
-	if err := netlink.LinkSetDown(vrf); err != nil {
+	if err := netlink.LinkSetDown(vlan); err != nil {
```
venkatmahalingam, line 175: Any reason to bring the SVI down before deletion?

pkg/evpn/svi.go
```diff
@@ -213,7 +246,7 @@ func (s *Server) GetTunnel(_ context.Context, in *pb.GetTunnelRequest) (*pb.Tunn
 		return nil, err
```
venkatmahalingam, line 246: We could have an util funtion for initial validation instead of repeating the same the every SVI create/update/del function.

pkg/evpn/svi.go
glimchb: I thought this is more proper, but you are the expert

pkg/evpn/svi.go
glimchb: yes

glimchb: done

pkg/evpn/svi.go
glimchb: for sure, will start refactoring after this one, after switch is completed

pkg/evpn/svi.go
```diff
@@ -27,88 +27,122 @@ import (
...
+		fmt.Printf("Failed to add vlan to bridge: %v", err)
+		return nil, err
+	}
+	// Example: ip link add link br-tenant name <link_svi> type vlan id <vlan-id>
+	vlandev := &netlink.Vlan{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:        resourceID,
+			ParentIndex: bridge.Attrs().Index,
+		},
+		VlanId: int(bridgeObject.Spec.VlanId),
```
mardim91, line 87: How this code snippet is equivalent to the 
ip link add link br-tenant name <link_svi> type vlan id <vlan-id> 
command I do not see the realtion. From what I understnad you just create a netlink.Vlan object but you do not execute any netlink function that can justify the command in the example. no ?

pkg/evpn/svi.go
mardim91: No harm to bring it down first. I will agree with @glimchb

pkg/evpn/svi.go
```diff
@@ -130,51 +164,51 @@ func (s *Server) DeleteTunnel(_ context.Context, in *pb.DeleteTunnelRequest) (*e
 		return nil, err
 	}
 	resourceID := path.Base(obj.Name)
-	// use netlink to find Subnet/Bridge
-	vrf, err := netlink.LinkByName(resourceID)
+	// use netlink to find vlan
+	vlan, err := netlink.LinkByName(resourceID)
```
mardim91, line 168: maybe vlandev or svidev would be more proper because you actually delete a vlan subinterface not just a vlan

mardim91 (changes requested):

pkg/evpn/svi.go
glimchb: You are absolutely correct, thanks for catching this. My bad. Was in the hurry and missed it

Issue #83, opened 2023-08-02
`CreateLogicalBridge` what about hard coded IP `10.0.0.100` ?


glimchb:
I see no field for IP address
```
ip link add vni10 type vxlan id 10 local 10.0.0.100 dstport 4789 nolearning proxy
ip link set vni10 master br-tenant up
bridge vlan add dev vni10 vid 10 pvid untagged
bridge link set dev vni10 neigh_suppress on
```
but
```
message LogicalBridgeSpec {
    // the VLAN of the L2 domain
    uint32 vlan_id         = 1 [(google.api.field_behavior) = REQUIRED];
    //VXLAN VNI for the L2 EVPN. Also used as EVPN route target
    uint32 vni             = 2;
}
```
@mardim91 please advise

mardim91:
Yes @glimchb as I have explained before maybe this is something that is missing from the API as the VRF spec have VTEP IP but the LogicalBridgeSpec it doesn't. Maybe we need to add this VTEP IP field to the API or completely remove it also from the VRF spec and pass it in both cases through evpn-gw-br configuration files. I will need to discuss this with @JanScheurich before we do any changes to the API. For now when it comes to the LogicalBridges you can pass the VTEP IP through configuration in evpn-gw-br.

glimchb:
thanks @mardim91 I will leave this issue open for tracking

glimchb:
@JanScheurich please advise

Issue #75, opened 2023-07-24
Create loopback interface and use /32 length instead 24


glimchb:
The loopback IP I believe can be /32 length. Is there any reason to use /24 lentgh ? @venkatmahalingam WDYT ?

and

              Yes, loopback interface IP should be /32, but we need to create a loopback interface and assign the IP, should be assigned on the VRF interface.

_Originally posted by @mardim91 in https://github.com/opiproject/opi-evpn-bridge/pull/66#discussion_r1271947746_

glimchb:
from @JanScheurich also `Also the vtep_ip_prefix must be /32`

PR #103, opened 2023-08-09
ci: replace static comamnds with grpc calls

glimchb:
- used https://www.browserling.com/tools/ip-to-dec to convert IP string to number
- used `python -c "import base64; print(base64.b64encode('\xaa\xbb\xcc\x00\x00\x42'))"` for MAC translation


docker-compose.yml
```diff
@@ -54,93 +54,6 @@ services:
...
-            ip link set eth1 up && ip addr flush dev eth1 && \
-            ip link set eth1 master br-tenant up && \
-            bridge vlan add dev eth1 vid 10 && \
-
-            bridge vlan add dev eth1 vid 20 && \
-
-            bridge vlan add dev eth1 vid 40 && \
-
-            ip addr flush dev eth2 && ip link set eth2 master br-tenant up && \
-            bridge vlan add dev eth2 vid 50 && \
```
glimchb, line 142: I assumed `eth2` is `ACCESS` and added `pvid untagged`, correct or not ?

docker-compose.yml
```diff
@@ -334,82 +247,35 @@ services:
...
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge    "{\"logical_bridge\" : {\"spec\" : {\"vni\" : 33, \"vlan_id\": 33 }}, \"logical_bridge_id\" : \"vxlan33\" }" && \
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateBridgePort    "{\"bridge_port\" : {\"spec\" : {\"mac_address\" : \"qrvMAAAB\", \"ptype\": \"ACCESS\", \"logical_bridges\": [] }}, \"bridge_port_id\" : \"eth2\" }" && \
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateBridgePort    "{\"bridge_port\" : {\"spec\" : {\"mac_address\" : \"qrvMAAAB\", \"ptype\": \"TRUNK\" , \"logical_bridges\": [] }}, \"bridge_port_id\" : \"eth3\" }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"blue\",     \"vrf\" :             {\"spec\" : {\"vni\" : 100, \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
```
glimchb, line 250: `vrf blue` create here will also create bridge `br100` and vxlan `vni100` which was not in the original config

docker-compose.yml
```diff
@@ -334,82 +247,35 @@ services:
...
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni10\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 10, \"vlan_id\": 10 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni20\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 20, \"vlan_id\": 20 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni30\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 30, \"vlan_id\": 30 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni40\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 40, \"vlan_id\": 40 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni50\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 50, \"vlan_id\": 50 }} }" && \
```
glimchb, line 257: `vni50` is not in the static config, but I need it to create SVI `vlan50`... what am I missing ?

docker-compose.yml
glimchb: fixed using #106

docker-compose.yml
glimchb: opened #104 for tracking

docker-compose.yml
glimchb: fixed using #98

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
...
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni30\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 30, \"vlan_id\": 30 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni40\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 40, \"vlan_id\": 40 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni50\",    \"logical_bridge\" :  {\"spec\" : {              \"vlan_id\": 50 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateSvi            "{\"svi_id\"            : \"vlan20\",   \"svi\" :             {\"spec\" : {\"vrf\": \"//network.opiproject.org/vrfs/blue\",   \"logical_bridge\": \"//network.opiproject.org/bridges/vni20\", \"mac_address\" : \"qrvMAAAh\", \"gw_ip_prefix\": [{\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 336860161}, \"len\": 24}] }} }" && \
```
JanScheurich, line 258: I would prefer to name the SVI objects "<VRF>-<LB>" so that the association between VRF and logical bridge that they create is clear already from the name. The id of the SVI object on the gRPC API does not necessarily have to be the same as the interface name created in Linux. The latter is an implementation detail.

JanScheurich (changes requested):

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
       ip link add br-tenant type bridge vlan_default_pvid 0 vlan_filtering 1 vlan_protocol 802.1Q && \
       ip link set br-tenant up && \
       echo create && \
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf    "{\"vrf\" : {\"spec\" : {\"vni\" : 100, \"loopback_ip_prefix\": {}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }}, \"vrf_id\" : \"blue\" }" && \
```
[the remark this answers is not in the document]
glimchb: @JanScheurich nice catch on the ip addresses, will fix now
regarding decimal, in real EVPN CLI here we do use strings https://github.com/opiproject/godpu/blob/32aa38d15fbe78cb7bea22fb04ae780606e51f0b/cmd/evpn-vrf.go#L47
in this docker-compose for now we using grpcli still, so this is temporal inconvenience

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
...
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"green\",    \"vrf\" :             {\"spec\" : {\"vni\" : 100, \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"yellow\",   \"vrf\" :             {\"spec\" : {\"vni\" : 101, \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vni10\",    \"logical_bridge\" :  {\"spec\" : {\"vni\" : 10, \"vlan_id\": 10 }} }" && \
```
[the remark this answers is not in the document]
glimchb: looks like this comment is related to the original iproute2 commands, since here I'm using the same name

docker-compose.yml
JanScheurich: Sorry, my comment got screwed up by the md interpreter. An SVI should be called e.g. `green-vlan10` if it connects VRF  `green` to logical bridge `vlan10`.

glimchb: thanks. You can use `back quotes` @JanScheurich to avoid formatting :-)

docker-compose.yml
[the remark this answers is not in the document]
glimchb: I agree that resource ID should be separated from linux device names
The intent of this PR for me was to recreate an exactly identical copy of the original static iproute2 docker compose script
Ci we can start using GRPC as soon as possible
So all the changes we make on top of this will still be protected by CI/CD pipeline to verify everything is still working as expected

docker-compose.yml
glimchb: sure

docker-compose.yml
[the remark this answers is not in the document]
glimchb: fixed in 1ca76e117428cc73510947ad8bd690439fca2c71

docker-compose.yml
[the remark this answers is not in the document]
glimchb: fixed in 4ff58656483d858cda2c6c0184cd1f4c5bf0883c

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
...
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge    "{\"logical_bridge\" : {\"spec\" : {\"vni\" : 33, \"vlan_id\": 33 }}, \"logical_bridge_id\" : \"vxlan33\" }" && \
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateBridgePort    "{\"bridge_port\" : {\"spec\" : {\"mac_address\" : \"qrvMAAAB\", \"ptype\": \"ACCESS\", \"logical_bridges\": [] }}, \"bridge_port_id\" : \"eth2\" }" && \
-      /entrypoint.sh call --json_input --json_output localhost:50151 CreateBridgePort    "{\"bridge_port\" : {\"spec\" : {\"mac_address\" : \"qrvMAAAB\", \"ptype\": \"TRUNK\" , \"logical_bridges\": [] }}, \"bridge_port_id\" : \"eth3\" }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"blue\",          \"vrf\" :             {\"spec\" : {               \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772418}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
```
venkatmahalingam, line 250: Prefix-len for loopback address should be 32, right?

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
...
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vlan30\",        \"logical_bridge\" :  {\"spec\" : {\"vni\" : 30, \"vlan_id\": 30 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vlan40\",        \"logical_bridge\" :  {\"spec\" : {\"vni\" : 40, \"vlan_id\": 40 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vlan50\",        \"logical_bridge\" :  {\"spec\" : {              \"vlan_id\": 50 }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateSvi            "{\"svi_id\"            : \"blue-vlan20\",   \"svi\" :             {\"spec\" : {\"vrf\": \"//network.opiproject.org/vrfs/blue\",   \"logical_bridge\": \"//network.opiproject.org/bridges/vlan20\", \"mac_address\" : \"qrvMAAAh\", \"gw_ip_prefix\": [{\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 336860161}, \"len\": 24}] }} }" && \
```
venkatmahalingam, line 258: Can we use vlan20 instead of blue-vlan20 for SVI name? and logical_bridge_id as br<vlan-id>? ,e.g br50

docker-compose.yml
```diff
@@ -54,93 +54,6 @@ services:
...
-
-            ip link add vni30 type vxlan id 30 local 10.0.0.2 dstport 4789 nolearning && \
-            ip link set vni30 master br-tenant up && \
-            bridge vlan add dev vni30 vid 30 pvid untagged && \
-            bridge link set dev vni30 neigh_suppress on learning off && \
-
-            ip link add vni40 type vxlan id 40 local 10.0.0.2 dstport 4789 nolearning && \
-            ip link set vni40 master br-tenant up && \
-            bridge vlan add dev vni40 vid 40 pvid untagged && \
-            bridge link set dev vni40 neigh_suppress on learning off && \
```
venkatmahalingam, line 107: How the fields "neigh_suppress" and "learning off" are mapped to the API?

docker-compose.yml
```diff
@@ -334,23 +247,35 @@ services:
...
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"green\",         \"vrf\" :             {\"spec\" : {\"vni\" : 100, \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772674}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateVrf            "{\"vrf_id\"            : \"yellow\",        \"vrf\" :             {\"spec\" : {\"vni\" : 101, \"loopback_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772930}, \"len\": 24}, \"vtep_ip_prefix\": {\"addr\": {\"af\": \"IP_AF_INET\", \"v4_addr\": 167772162}, \"len\": 24} }} }" && \
+      /entrypoint.sh call --json_input --json_output localhost:50151 CreateLogicalBridge  "{\"logical_bridge_id\" : \"vlan10\",        \"logical_bridge\" :  {\"spec\" : {\"vni\" : 10, \"vlan_id\": 10 }} }" && \
```
venkatmahalingam, line 253: ip link add vni10 type vxlan id 10 local 10.0.0.2 dstport 4789 nolearning

How's the above command is mapped in the API? where to provide VTEP IP in the API?

docker-compose.yml
JanScheurich: I specifically want the SVI name to reflect the binding between VRF and logical bridge.
And I don't see any reason to encode the object type as prefix into the logical bridge objects. We are not doing that for other object types either. How we call the underlying Linux realization entities is a different matter.

docker-compose.yml
JanScheurich: They don't have to be reflected in the API. They are hard-coded in the EVPN-bridge implementation (until the point where someone shows that it is necessary to be able to control them. Then we can discuss how to enhance the logical bridge API with an optional attribute).

docker-compose.yml
JanScheurich: The VRF spec has a field `vtep_ip_prefix` for this purpose.

docker-compose.yml
JanScheurich: Also the `vtep_ip_prefix` must be `/32`

docker-compose.yml
venkatmahalingam: Why do we need a VRF for L2VXLAN case?

docker-compose.yml
venkatmahalingam: How we call the underlying Linux realization entities is a different matter.
-> This is the problem, unnecessary lengthy names lead to have additional mapping in the backend as Linux interface names length restricted to 15 characters. IMO, it is easy to get and know the SVI to VRF mappings, if we want to rebind the SVI from one VRF to another VRF on the fly, I would need to delete the SVI and create with new VRF name part of SVI name.

JanScheurich (approved):
LGTM

docker-compose.yml
JanScheurich: Good catch, the LogicalBridgeSpec in the API should also have a `vtep_ip_prefix`. See issue #83

docker-compose.yml
glimchb: I agree with @venkatmahalingam on this problem
```
if we want to rebind the SVI from one VRF to another VRF on the fly, I would need to delete the SVI and create it with new VRF name part of SVI name.
```
But naming of objects is left to the user, there is no code or rules to mandate to use any names... so we as OPI should not care... 
user will pick a name or  (since resource ID) is optional, grpc server will generate UUID as a name (or more accurately resource ID) instead...

docker-compose.yml
glimchb: opened #75 and create a new PR, since need some code change to accommodate with this

docker-compose.yml
glimchb: opened #147

docker-compose.yml
JanScheurich: How realistic and important a use case is it to move a subnet on a logical bridge to another VRF without any traffic impact? To me that is a change in network planning, which typically has wider implications in the network.
I know it is possible in Linux to shift an existing SVI netdev to another VRF just by enslaving it to the new vrf netdev, but is this something the EVPN-GW API must cover? In fact, I think it currently does cover it, as the VRF reference in the SVI spec is not marked as `immutable`. Same goes for the LogicalBridge reference, by the way, but the VLAN ID of an SVI netdev can't be changed in Linux on the fly. 
At any rate, the object naming arbitrary, and we can apply any convention we like in this example configuration.

PR #107, opened 2023-08-11
Introduce infraDB for persistent store of the objects

mardim91:


glimchb (changes requested):
for me this is not providing the right abstraction using repository design pattern
tightly coupling specific DB implementation inside the code
I will post alternative PR how this should be done properly and update link in comments

mardim91:
> for me this is not providing the right abstraction using repository design pattern tightly coupling specific DB implementation inside the code I will post alternative PR how this should be done properly and update link in comments

Can you please share more information on what is the problem with this approach and what would you  like to change ? The reason I am saying this is because this architecture will fit well also to the later changes that are coming where the frr and kernel implementation will be separated to their own modules which they will receive notifications from database through a pub/sub channel when we have updates of objects in the database.

pkg/evpn/vrf.go
```diff
@@ -18,7 +18,9 @@ import (
 	"github.com/google/uuid"
 	"github.com/vishvananda/netlink"
 
+	badger "github.com/dgraph-io/badger/v4"
```
glimchb, line 21: try to avoid importing `badger` directly here 
you using it mostly for error checking, you should translate badger errors to infradb errors and hide the fact we actually using badger in this file... only main should know about implementation details like this

pkg/infradb/manager.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type VrfManager struct {
+	VrfObj *pb.Vrf
+}
+
+func (v *VrfManager) Add() error {
+
+	err := db.Update(func(txn *badger.Txn) error {
+		value, err := json.Marshal(v.VrfObj)
```
glimchb, line 47: according to https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson
```
This package produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.
```
so please use
```
import "google.golang.org/protobuf/proto"
value, err := proto.Marshal(v.VrfObj)
```
instead

pkg/evpn/vrf.go
mardim91: I will fix that

pkg/infradb/manager.go
mardim91: I will fix that

docs/diagrams/CreateBP.puml
```diff
@@ -0,0 +1,59 @@
+@startuml
```
glimchb, line 1: if you use [mermaid ](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/) then github will render the picture for you as well, which is very nice, please consider

mardim91: Thank you for letting me know. I will probably will

mardim91:
I will close this PR as it will be replaced by a new PR in the future

PR #121, opened 2023-08-14
feat(vrf): delete bridge and vxlan netdevs

### LINKED ISSUE #120: vrf: in `delete` function, also remove bridge and vxlan devices
see https://github.com/opiproject/opi-evpn-bridge/blob/e8c2c86890ce67ab7d1cc7568d977250a79b3a9b/pkg/evpn/vrf.go#L184

glimchb:
Fixes #120

### DIFF
--- a/pkg/evpn/vrf.go
+++ b/pkg/evpn/vrf.go
@@ -166,6 +166,47 @@ func (s *Server) DeleteVrf(_ context.Context, in *pb.DeleteVrfRequest) (*emptypb
 		log.Printf("error: %v", err)
 		return nil, err
 	}
+	// delete bridge and vxlan only if VNI value is not empty
+	if obj.Spec.Vni != nil {
+		// use netlink to find VXLAN device
+		vxlanName := fmt.Sprintf("vni%d", *obj.Spec.Vni)
+		vxlandev, err := netlink.LinkByName(vxlanName)
+		log.Printf("Deleting VXLAN %v", vxlandev)
+		if err != nil {
+			err := status.Errorf(codes.NotFound, "unable to find key %s", vxlanName)
+			log.Printf("error: %v", err)
+			return nil, err
+		}
+		// bring link down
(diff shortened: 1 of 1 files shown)


pkg/evpn/vrf.go
```diff
@@ -166,6 +166,47 @@ func (s *Server) DeleteVrf(_ context.Context, in *pb.DeleteVrfRequest) (*emptypb
...
+			return nil, err
+		}
+		// use netlink to delete VXLAN device
+		if err := netlink.LinkDel(vxlandev); err != nil {
+			fmt.Printf("Failed to delete link: %v", err)
+			return nil, err
+		}
+		// use netlink to find BRIDGE device
+		bridgeName := fmt.Sprintf("br%d", *obj.Spec.Vni)
+		bridgedev, err := netlink.LinkByName(bridgeName)
```
venkatmahalingam, line 192: Hope we could have stored in the DB instead of querying it during deletion.

glimchb: DB is not helping here much...

`netlink` library works on link object, see https://github.com/vishvananda/netlink/blob/6afddb37c1f00693528264222a57a3c08dccbab7/link_linux.go#L1700

- so either we have to call `netlink.LinkByName(bridgeName)` to get the object handle...
```
bridgedev, err := netlink.LinkByName(bridgeName)
netlink.LinkDel(bridgedev)
```
- The other alternative (**didn't test it yet**) is to create new object with the same name and then delete it
```
bridgedev := &netlink.Bridge{LinkAttrs: netlink.LinkAttrs{Name: bridgeName}}
netlink.LinkDel(bridgedev)
```

venkatmahalingam (approved):

pkg/evpn/vrf.go
glimchb: opened #123 for tracking

PR #148, opened 2023-08-31
fix(bridge): avoid LinkByName netlink calls

### LINKED ISSUE #123: consider avoid `LinkByName` for performance
`netlink` library works on link object, see https://github.com/vishvananda/netlink/blob/6afddb37c1f00693528264222a57a3c08dccbab7/link_linux.go#L1700

consider instead of
```
bridgedev, err := netlink.LinkByName(bridgeName)
netlink.LinkDel(bridgedev)
```
do
```
bridgedev := &netlink.Bridge{LinkAttrs: netlink.LinkAttrs{Name: bridgeName}}
netlink.LinkDel(bridgedev)
```
maybe...

glimchb:
Main reason is perfoamnce.
If we have the relevant info in the cache/DB,
we could just use it instead of netlink get calls.

Fixes #123

### DIFF
--- a/pkg/evpn/bridge.go
+++ b/pkg/evpn/bridge.go
@@ -67,12 +67,6 @@ func (s *Server) CreateLogicalBridge(_ context.Context, in *pb.CreateLogicalBrid
 	}
 	// create vxlan only if VNI is not empty
 	if in.LogicalBridge.Spec.Vni != nil {
-		bridge, err := s.nLink.LinkByName(tenantbridgeName)
-		if err != nil {
-			err := status.Errorf(codes.NotFound, "unable to find key %s", tenantbridgeName)
-			log.Printf("error: %v", err)
-			return nil, err
-		}
 		// Example: ip link add vxlan-<LB-vlan-id> type vxlan id <LB-vni> local <vtep-ip> dstport 4789 nolearning proxy
 		myip := make(net.IP, 4)
 		binary.BigEndian.PutUint32(myip, in.LogicalBridge.Spec.VtepIpPrefix.Addr.GetV4Addr())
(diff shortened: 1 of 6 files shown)


venkatmahalingam (approved):

mardim91 (commented):
Some comments
1. I do not see where the tenantbridgeName variable gets assigned (take its value).
2. I am not sure if it is the same thing creating a link object by hand compared to returning the link object by using the linkByName function. Maybe the returned object has more attributes that are needed during the subsequent netlink call (just a speculation).
3. I prefer the LinkByName as this way we are sure that object actually exists on the system and has not been deleted accidentally by a third party process.
4. I do not believe has a huge performance overhead. WDYT ?

venkatmahalingam:
> Some comments
> 
> 1. I do not see where the tenantbridgeName variable gets assigned (take its value).
> 2. I am not sure if it is the same thing creating a link object by hand compared to returning the link object by using the linkByName function. Maybe the returned object has more attributes that are needed during the subsequent netlink call (just a speculation).
> 3. I prefer the LinkByName as this way we are sure that object actually exists on the system and has not been deleted accidentally by a third party process.
> 4. I do not believe has a huge performance overhead. WDYT ?

I thought querying a Linux kernel every time while we maintain a cache is unnecessary, if there is a possibility of third party process deletion, IMO, that's not a good design/coding practice and we should have a single source of truth.

mardim91:
> > Some comments
> > 
> > 1. I do not see where the tenantbridgeName variable gets assigned (take its value).
> > 2. I am not sure if it is the same thing creating a link object by hand compared to returning the link object by using the linkByName function. Maybe the returned object has more attributes that are needed during the subsequent netlink call (just a speculation).
> > 3. I prefer the LinkByName as this way we are sure that object actually exists on the system and has not been deleted accidentally by a third party process.
> > 4. I do not believe has a huge performance overhead. WDYT ?
> 
> I thought querying a Linux kernel every time while we maintain a cache is unnecessary, if there is a possibility of third party process deletion, IMO, that's not a good design/coding practice and we should have a single source of truth.

Not really sure. To be honest I think is more proper to for me actually get the link object that is returned and use it for deletion than creating a new object with just the name. Also we do no know how the subsequent netlink calls will behave if we do no actually get the real object but we create one by hand. Maybe they will nee some attributed from the object that do no exist and that make them fail.

Also kubernetes CNIs use also the LinkByName approach in order to do further netlink operations check[ here ](https://github.com/k8snetworkplumbingwg/sriov-cni/blob/master/pkg/sriov/sriov.go#L128)

PR #155, opened 2023-09-06
test: use mock.Anything to simplify tests

glimchb:


glimchb:
simplifying tests like this will cause them to be more weak, so deciding to abandon this change

PR #170, opened 2023-09-08
feat(db): add repository package for persistence

### LINKED ISSUE #178: feat(store): use gokv pkg to abstract persistant store
see https://github.com/opiproject/opi-evpn-bridge/pull/177

glimchb:
This is not full implementation yet, just the start

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -13,6 +13,7 @@ import (
 
 	pe "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
 	"github.com/opiproject/opi-evpn-bridge/pkg/evpn"
+	"github.com/opiproject/opi-evpn-bridge/pkg/repository"
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/reflection"
@@ -28,6 +29,17 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to listen: %v", err)
 	}
+	// Provision for persistent storage here
+	dbvrf, err := repository.VrfFactory("memory")
+	if err != nil {
+		log.Fatalf("failed to create database: %v", err)
+	}
+	dbvsvi, err := repository.SviFactory("memory")
+	if err != nil {
+		log.Fatalf("failed to create database: %v", err)
+	}
+	log.Printf("todo: use db in opi server (%v) and (%v)", dbvrf, dbvsvi)
+
 	s := grpc.NewServer()
 	opi := evpn.NewServer()
 
--- a/pkg/repository/models_svi.go
+++ b/pkg/repository/models_svi.go
@@ -1,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Dell Inc, or its subsidiaries.
+
+// Package repository is the database abstraction implementing repository design pattern
+package repository
+
+import (
+	"encoding/binary"
+	"net"
+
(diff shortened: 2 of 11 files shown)


pkg/repository/repository.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Database abstraction
+type Database interface {
+	Set(key string, value string) (string, error)
+	Get(key string) (string, error)
+	Delete(key string) (string, error)
+}
+
+// Factory pattern to create new Database
+func Factory(databaseImplementation string) (Database, error) {
```
pwalessi-dell, line 15: You probably want to define a database enum instead of using a string

pkg/repository/repository.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Dell Inc, or its subsidiaries.
+
+// Package repository is the database abstraction implementing repository design pattern
+package repository
+
+// Database abstraction
+type Database interface {
```
pwalessi-dell, line 8: You might want to call it KVStore or something like that. It's not really a "database".

pwalessi-dell (approved):
I think that this is good! Just a couple of comments, no biggies though.

cmd/main.go
```diff
@@ -28,6 +29,12 @@ func main() {
 	if err != nil {
 		log.Fatalf("failed to listen: %v", err)
 	}
+	db, err := repository.Factory("memory")
```
brianroch, line 32: I assume the `Database` interface (or `db` instance here) would be passed to subsequent services, right? Allowing you to pass a mock database to those services during testing.

pkg/repository/repository.go
brianroch: You might also consider including mocked implementations of this interface to encourage that testing pattern.

pkg/repository/repository.go
brianroch: For databases that are backed by a provided service, i.e. Redis (or anything not in memory or a persistent volume), you will need to pass the url, host, port, user, pass, etc... I think it would be worth creating a `config` struct in the `repository` package for these values and passing that to the factory too. All keys would be optional and you would rely on code documentation to state which keys are required for each implementation.

You could include the `databaseImplementation` value as part of this `config` struct too.

Depending on the flexibility required, you could create a `config` struct based on a free-form `map[string]string`. This wouldn't force you to pre-define the keys required for a `Database` instance, but instead you would rely on database implementations clearly documenting the required key-value config entries required. An example of this is here: https://matthewbrown.io/2016/01/23/factory-pattern-in-golang

cmd/main.go
glimchb: correct

pkg/repository/models_svi.go
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Dell Inc, or its subsidiaries.
```
glimchb, line 2: moved to a separate PR #267

pkg/repository/models_vrf.go
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Dell Inc, or its subsidiaries.
+
+// Package repository is the database abstraction implementing repository design pattern
```
glimchb, line 4: moved to a separate PR #267

PR #177, opened 2023-09-12
feat(store): replace all 4 objects maps with gokv.Store abstraction

glimchb:
see https://github.com/philippgille/gokv

### DIFF
--- a/pkg/evpn/evpn.go
+++ b/pkg/evpn/evpn.go
@@ -37,7 +37,6 @@ type Server struct {
 	Bridges    map[string]*pb.LogicalBridge
 	Ports      map[string]*pb.BridgePort
 	Svis       map[string]*pb.Svi
-	Vrfs       map[string]*pb.Vrf
 	Pagination map[string]int
 	nLink      utils.Netlink
 	frr        utils.Frr
@@ -68,7 +67,6 @@ func NewServerWithArgs(nLink utils.Netlink, frr utils.Frr, store gokv.Store) *Se
 		Bridges:    make(map[string]*pb.LogicalBridge),
 		Ports:      make(map[string]*pb.BridgePort),
 		Svis:       make(map[string]*pb.Svi),
-		Vrfs:       make(map[string]*pb.Vrf),
 		Pagination: make(map[string]int),
 		nLink:      nLink,
 		frr:        frr,
--- a/pkg/evpn/evpn.go
+++ b/pkg/evpn/evpn.go
@@ -35,6 +35,7 @@ type Server struct {
 	pb.UnimplementedLogicalBridgeServiceServer
 	pb.UnimplementedBridgePortServiceServer
 	Pagination map[string]int
+	ListHelper map[string]bool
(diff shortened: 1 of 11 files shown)


artek-koltun:
@glimchb I think that the chosen library `gokv` have some disadvantages: It is not updated since 2020. It contains only 11 commits, 0 forks and 0 stars. So there is no active community maintaining it. This entails the risks: What if we have a bug there, or we need some additional feature? Who would do that work? Add/fix the functionality on our own? But I suppose we need a library to offload that job
I found [gofiber/storage](https://github.com/gofiber/storage/tree/main) project. It is maintained. It has 211 stars. Furthermore, it has a discord channel, so we can reach out for help if needed. In addition, it supports a broader set of storage backends. Would it be better to depend on that library instead of `gokv`?

pkg/evpn/port_test.go
```diff
@@ -452,7 +457,10 @@ func Test_DeleteBridgePort(t *testing.T) {
 			// start GRPC mockup server
 			ctx := context.Background()
 			mockNetlink := mocks.NewNetlink(t)
-			opi := NewServerWithArgs(mockNetlink)
+			options := gomap.DefaultOptions
+			options.Codec = utils.ProtoCodec{}
+			store := gomap.NewStore(options)
+			opi := NewServerWithArgs(mockNetlink, store)
```
brianroch, line 463: Bit of repetition here, could this be extracted to a function?

artek-koltun:
> @glimchb I think that the chosen library `gokv` have some disadvantages: It is not updated since 2020. It contains only 11 commits, 0 forks and 0 stars. So there is no active community maintaining it. This entails the risks: What if we have a bug there, or we need some additional feature? Who would do that work? Add/fix the functionality on our own? But I suppose we need a library to offload that job I found [gofiber/storage](https://github.com/gofiber/storage/tree/main) project. It is maintained. It has 211 stars. Furthermore, it has a discord channel, so we can reach out for help if needed. In addition, it supports a broader set of storage backends. Would it be better to depend on that library instead of `gokv`?

ok, my oversight. messed up with repository and referred to https://github.com/yifeng01/gokv
https://github.com/philippgille/gokv repo looks ok

pkg/evpn/evpn.go
```diff
@@ -34,11 +34,8 @@ type Server struct {
 	pb.UnimplementedSviServiceServer
 	pb.UnimplementedLogicalBridgeServiceServer
 	pb.UnimplementedBridgePortServiceServer
-	Bridges    map[string]*pb.LogicalBridge
-	Ports      map[string]*pb.BridgePort
-	Svis       map[string]*pb.Svi
-	Vrfs       map[string]*pb.Vrf
 	Pagination map[string]int
+	ListHelper map[string]bool
```
artek-koltun, line 38: @glimchb since `ListHelper` is used as a set, it is more optimal to use `struct{}` as a value to save up 1 byte per entry (empty struct occupies 0 bytes).

PR #246, opened 2023-10-14
feat(frr): replace static config with dynamic grpc

glimchb:

### DIFF
--- a/pkg/evpn/vrf.go
+++ b/pkg/evpn/vrf.go
@@ -140,7 +140,8 @@ func (s *Server) CreateVrf(ctx context.Context, in *pb.CreateVrfRequest) (*pb.Vr
 	}
 	// configure FRR
 	if err := s.frrCreateVrfRequest(ctx, in); err != nil {
-		return nil, err
+		fmt.Printf("skip err check for now: %v", err)
+		// return nil, err
 	}
 	// save object to the database
 	response := protoClone(in.Vrf)
@@ -223,7 +224,8 @@ func (s *Server) DeleteVrf(ctx context.Context, in *pb.DeleteVrfRequest) (*empty
 	}
 	// delete from FRR
 	if err := s.frrDeleteVrfRequest(ctx, obj); err != nil {
-		return nil, err
+		fmt.Printf("skip err check for now: %v", err)
+		// return nil, err
 	}
 	// remove from the Database
 	delete(s.Vrfs, obj.Name)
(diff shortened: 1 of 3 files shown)


pkg/evpn/vrf.go
```diff
@@ -141,11 +141,31 @@ func (s *Server) CreateVrf(ctx context.Context, in *pb.CreateVrfRequest) (*pb.Vr
...
-			vrf %s-opi
-				vni 3%d
+			vrf %s
+				vni %d
 				exit-vrf
 			exit`, vrfName, *in.Vrf.Spec.Vni))
 		fmt.Printf("FrrZebraCmd: %v:%v", data, err)
+		data1, err1 := utils.FrrBgpCmd(ctx, fmt.Sprintf(
+			`configure terminal
+			router bgp 65000 vrf %s
```
venkatmahalingam, line 151: What gRPC config for BGP is expected from the user? this looks like we hardcode all the configs inside.

mardim91: The BGP config is out of scope right now. We care mostly about those minimal parameters that will come from the user that will make the EVPN work. So most of the BGP config is static right now.

mardim91 (commented):
Do we have any implementation for these parameters ?

https://github.com/opiproject/opi-api/blob/main/network/evpn-gw/v1alpha1/l3_xpu_infra_mgr.proto#L274-L278

Because when those parameters are passed we need to have extra config into the FRR when an SVI is created with the above parameters enabled. The extra confi looks like this


vtysh \
        -c "conf t" \
        -c "router bgp <-as-number-> vrf <-vrf-name->" \
        -c "  bgp disable-ebgp-connected-route-check" \
        -c "  neighbor <-svi-name-> peer-group" \
        -c "  neighbor <-svi-name-> remote-as <-remote-as-number->" \
        -c "  neighbor <-svi-name-> update-source <-svi-ip-adress->" \
        -c "  neighbor <-svi-name-> as-override" \
        -c "  neighbor <-svi-name-> soft-reconfiguration inbound" \
        -c "  bgp listen range <-svi-ip-adress-with prefix-length-> peer-group <-svi-name->" \
        -c "exit"

glimchb:
> Do we have any implementation for these parameters ?

Yes, I do.
Once we merge this PR for `CreateVrf` I will open new PR for `CreateSvi`
make sense ?

glimchb:
> Do we have any implementation for these parameters ?
> https://github.com/opiproject/opi-api/blob/main/network/evpn-gw/v1alpha1/l3_xpu_infra_mgr.proto#L274-L278

started to work on them here #252

PR #267, opened 2023-10-18
feat(models): add new package to handle protobuf translation

glimchb:

This is just a first draft and will be amended later on to include more fields and fix Todos...

There are 4 objects:
- [`LogicalBridgeSpec`](https://github.com/opiproject/opi-api/blob/d81cc5ee60d4df53cf660ced62e8fe5b17de4cfd/network/evpn-gw/v1alpha1/l2_xpu_infra_mgr.proto#L132-L144)
- [`BridgePortSpec`](https://github.com/opiproject/opi-api/blob/d81cc5ee60d4df53cf660ced62e8fe5b17de4cfd/network/evpn-gw/v1alpha1/l2_xpu_infra_mgr.proto#L238-L246)
- [`SviSpec`](https://github.com/opiproject/opi-api/blob/d81cc5ee60d4df53cf660ced62e8fe5b17de4cfd/network/evpn-gw/v1alpha1/l3_xpu_infra_mgr.proto#L256-L279)
- [`VrfSpec`](https://github.com/opiproject/opi-api/blob/d81cc5ee60d4df53cf660ced62e8fe5b17de4cfd/network/evpn-gw/v1alpha1/l3_xpu_infra_mgr.proto#L134-L143)

### DIFF
--- a/pkg/models/bridge.go
+++ b/pkg/models/bridge.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Dell Inc, or its subsidiaries.
+
+// Package models translates frontend protobuf messages to backend messages
+package models
+
+import (
+	// "encoding/binary"
+	"net"
+
+	pb "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+)
+
+// Bridge object, separate from protobuf for decoupling
+type Bridge struct {
+	Vni    uint32
+	VlanID uint32
+	VtepIP net.IPNet
+}
+
+// NewBridge creates new SVI object from protobuf message
+func NewBridge(in *pb.LogicalBridge) *Bridge {
+	// vtepip := make(net.IP, 4)
+	// binary.BigEndian.PutUint32(vtepip, in.Spec.VtepIpPrefix.Addr.GetV4Addr())
+	// vip := net.IPNet{IP: vtepip, Mask: net.CIDRMask(int(in.Spec.VtepIpPrefix.Len), 32)}
+	// TODO: Vni: *in.Spec.Vni
+	return &Bridge{VlanID: in.Spec.VlanId}
+}
+
+// ToPb transforms SVI object to protobuf message
+func (in *Bridge) ToPb() (*pb.LogicalBridge, error) {
+	bridge := &pb.LogicalBridge{
+		Spec: &pb.LogicalBridgeSpec{
+			Vni:    &in.Vni,
+			VlanId: in.VlanID,
+		},
+		Status: &pb.LogicalBridgeStatus{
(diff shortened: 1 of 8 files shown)


mardim91 (commented):
I think the changes are towards the right direction.

Is it possible instead of writing little code for all objects to just first write all the code or significant part of it just for one or two objects (e.g. VRF and SVI or LB and BP) so I can review the code better and be sure that I am agreeing with the translation approach ?

glimchb:
> I think the changes are towards the right direction.
> 
> Is it possible instead of writing little code for all objects to just first write all the code or significant part of it just for one or two objects (e.g. VRF and SVI or LB and BP) so I can review the code better and be sure that I am agreeing with the translation approach ?
> 
> That would really help me reviewing the code

This change defines interfaces. With new package, New and To functions and enough code to understand the new structure. The content of each function for specific transformation will come in subsequent PRs

pkg/models/bridge.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	// "encoding/binary"
+	"net"
+
+	pb "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+)
+
+// Bridge object, separate from protobuf for decoupling
+type Bridge struct {
```
glimchb, line 15: seems like we can also add interface here

pkg/models/svi.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"encoding/binary"
+	"net"
+
+	pb "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+)
+
+// Svi object, separate from protobuf for decoupling
+type Svi struct {
```
glimchb, line 15: seems like we can also add interface here

pkg/models/vrf.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"encoding/binary"
+	"net"
+
+	pb "github.com/opiproject/opi-api/network/evpn-gw/v1alpha1/gen/go"
+)
+
+// Vrf object, separate from protobuf for decoupling
+type Vrf struct {
```
glimchb, line 15: seems like we can also add interface here

brianroch:
Approach looks fine overall. The only alternative I think worth considering is using this [copier library](https://github.com/jinzhu/copier). It would save you a bit of boilerplate code by defining the field mapping in the struct itself. However, this would only work from pb to struct.

glimchb:
> Approach looks fine overall. The only alternative I think worth considering is using this [copier library](https://github.com/jinzhu/copier). It would save you a bit of boilerplate code by defining the field mapping in the struct itself. However, this would only work from pb to struct, not the other way.

opened #269 I was thinking also marshaling with json for auto-copy approach... will investigate this

PR #279, opened 2023-10-20
domain objects abstraction

brianroch:


glimchb:
can you please remove 
```
// TODO: move all of this to a common place
const (
	tenantbridgeName = "br-tenant"
)
```
from */common.go since we have it already in a common locations ?

pkg/objects/bridge/grpc.go
```diff
@@ -17,17 +17,16 @@ import (
...
-	"go.einride.tech/aip/fieldbehavior"
 	"go.einride.tech/aip/resourceid"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"google.golang.org/protobuf/types/known/emptypb"
 )
 
-// CreateLogicalBridge executes the creation of the LogicalBridge
-func (s *Server) CreateLogicalBridge(ctx context.Context, in *pb.CreateLogicalBridgeRequest) (*pb.LogicalBridge, error) {
+// Create executes the creation of the LogicalBridge
```
glimchb, line 25: if this function operates on models now, where in the code we no return grpc response ?

brianroch: Good point @glimchb. I've pulled back the abstraction to focus on consolidating how the operations are performed on the store. At the boundaries of each pb operation proto bufs are used, but the internal workings of those functions are consolidated to a set of store operations defined in `objects/object.go`

brianroch:
Let me know your thoughts on this approach @glimchb. Waiting on your feedback before proceeding.

PR #296, opened 2023-10-25
feat(frr): get frr address from external config

### LINKED ISSUE #294: Add frrAddress cli parameter
We need a way to configure frr address for the bridge, as we do [for example](https://github.com/opiproject/opi-evpn-bridge/blob/695ff86d9f174f387f2f1d2478bbed40fc5eab10/cmd/main.go#L49) for Redis.
After that, godpu docker-compose can be adjusted to use opi network for opi-evpn-bridge
https://github.com/opiproject/godpu/pull/351#discussion_r1371783082

glimchb:
Fixed #294 

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -49,6 +49,9 @@ func main() {
 	var redisAddress string
 	flag.StringVar(&redisAddress, "redis_addr", "127.0.0.1:6379", "Redis address in ip_address:port format")
 
+	var frrAddress string
+	flag.StringVar(&frrAddress, "frr_addr", "127.0.0.1", "Frr address in ip_address format, no port")
+
 	flag.Parse()
 
 	// Create KV store for persistence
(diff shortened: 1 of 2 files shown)


cmd/main.go
```diff
@@ -112,10 +115,10 @@ func runGrpcServer(grpcPort int, tlsFiles string, store gokv.Store) {
 	)
 	s := grpc.NewServer(serverOptions...)
 
-	bridgeServer := bridge.NewServer(store)
-	portServer := port.NewServer(store)
-	vrfServer := vrf.NewServer(store)
-	sviServer := svi.NewServer(store)
+	bridgeServer := bridge.NewServer(store, frrAddress)
```
artek-koltun, line 118: I prefer to inject dependency itself, rather than propagating a dependency parameter. If we need to add one more parameter, next time we do not pass it through the entire codebase, but set it in one place at wiring stage.

glimchb: I was thinking about it, we already have `NewServerWithArgs(store, nlink, frr)` so we can just use it...

artek-koltun (approved):

PR #308, opened 2023-11-02
test: add jaeger and gw tests to opi-test

artek-koltun:


glimchb (changes requested):
why not keep `opi-evpn-bridge` dependency ?

artek-koltun:
> why not keep `opi-evpn-bridge` dependency ?

I think opi-test should depend on only other tests. And if those tests need opi-evpn-server, they should specify it explicitly in their depends_on section.

glimchb:
> I think opi-test should depend on only other tests. And if those tests need opi-evpn-server, they should specify it explicitly in their depends_on section.

maybe... wasn't clear from this commit what your intentions are...

glimchb (approved):

glimchb (approved):

PR #380, opened 2024-04-29
fix(evpn-bridge): make tracer optional

mardim91:
Make tracer optional so we can import the netlinkWrapper library from evpn-gw-cni. 

More info here: https://github.com/opiproject/opi-gateway-evpn-cni/pull/14#discussion_r1515365216

### DIFF
--- a/pkg/utils/netlink.go
+++ b/pkg/utils/netlink.go
@@ -19,6 +19,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
 )
 
 // Netlink represents limited subset of functions from netlink package
@@ -73,7 +74,18 @@ type NetlinkWrapper struct {
 
 // NewNetlinkWrapper creates initialized instance of NetlinkWrapper
 func NewNetlinkWrapper() *NetlinkWrapper {
-	return &NetlinkWrapper{tracer: otel.Tracer("")}
+	return NewNetlinkWrapperWithArgs(true)
+}
+
+// NewNetlinkWrapperWithArgs creates initialized instance of NetlinkWrapper
+// based on passing arguments
+func NewNetlinkWrapperWithArgs(enableTracer bool) *NetlinkWrapper {
+	netlinkWrapper := &NetlinkWrapper{}
+	netlinkWrapper.tracer = noop.NewTracerProvider().Tracer("")
+	if enableTracer {
+		netlinkWrapper.tracer = otel.Tracer("")
+	}
+	return netlinkWrapper
 }
 
 // build time check that struct implements interface
--- a/pkg/utils/frr.go
+++ b/pkg/utils/frr.go
@@ -16,6 +16,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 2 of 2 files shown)


pkg/utils/netlink.go
```diff
@@ -35,6 +36,12 @@ type Netlink interface {
 	LinkSetDown(context.Context, netlink.Link) error
 	LinkSetMaster(context.Context, netlink.Link, netlink.Link) error
 	LinkSetNoMaster(context.Context, netlink.Link) error
+	LinkSetNsFd(context.Context, netlink.Link, int) error
```
artek-koltun, line 39: The PR descriptions says that we are making the tracer optional. However, we are adding new methods. Should it be a part of a separate commit?

pkg/utils/netlink.go
```diff
@@ -73,148 +80,261 @@ type NetlinkWrapper struct {
...
-	return &NetlinkWrapper{tracer: otel.Tracer("")}
+	return NewNetlinkWrapperWithArgs(true)
+}
+
+// NewNetlinkWrapperWithArgs creates initialized instance of NetlinkWrapper
+// based on passing arguments
+func NewNetlinkWrapperWithArgs(enableTracer bool) *NetlinkWrapper {
+	netlinkWrapper := &NetlinkWrapper{}
+	if enableTracer {
+		netlinkWrapper.tracer = otel.Tracer("")
```
artek-koltun, line 91: I see that we need to check a tracer all the time here and there...
Is it possible to use [NoopTracerProvider](https://pkg.go.dev/go.opentelemetry.io/otel/api/trace#NoopTracerProvider) here, so we use that noop object and not checking for nil all the time?

pkg/utils/netlink.go
mardim91: Hello @artek-koltun,

I will prefer to have those changes if it is possible to a single PR as it is easier for me to test the CNI [PR](https://github.com/opiproject/opi-gateway-evpn-cni/pull/14) against this one. Mainly these changes that are introduced in this PR have as purpose to make the netlink and FRR wrapper libraries to be reused by other components like the [EVPN GW CNI](https://github.com/opiproject/opi-gateway-evpn-cni/pull/14)

I can update the description of the PR if that is ok but I would prefer to not have multiple PRs as it is difficult to test the CNI against this. Hope that is ok.

pkg/utils/netlink.go
mardim91: That is a very good point. Let me take a look

pkg/utils/netlink.go
artek-koltun: What if you test locally, then split and provide one PR and then another one?

mardim91: ok let me do the changes on this PR and test them. If they are working I will discard this PR and create two individual ones.

mardim91: Hello @artek-koltun,

I have made the changes with the noop tracer and tested them. They are working. I will create the two individuals PRs tommorow as we have discussed above

mardim91: Hello @artek-koltun,

I have splitted the changes in two PRs. The first one is this PR here and the second one is [this](https://github.com/opiproject/opi-evpn-bridge/pull/383)

Please review when you find some time.

artek-koltun: Could I ask you to update the storage bridges as well?

artek-koltun (approved):
LGTM

mardim91:
Thank you @artek-koltun

PR #381, opened 2024-05-06
fix(evpn-bridge): fix empty vni use cases for vrf and logical bridge

atulpatel261194:


pkg/infradb/infradb.go
```diff
@@ -321,24 +321,25 @@ func UpdateLBStatus(name string, resourceVersion string, notificationID string,
...
 			if lb.Spec.Vni != nil {
-				delete(vpns, *lb.Spec.Vni)
-			}
-			err = infradb.client.Set("vpns", &vpns)
-			if err != nil {
-				log.Println(err)
-				return err
+				found, err = infradb.client.Get("vpns", &vpns)
+				if err != nil {
+					log.Println(err)
```
artek-koltun, line 327: is it a good strategy to log an error here and return it?
Apparently, a caller will likely also print the same error. Doing so, we are causing a lot of noise in the application logs for, in my opinion, little value.

pkg/infradb/infradb.go
```diff
@@ -975,24 +976,25 @@ func UpdateVrfStatus(name string, resourceVersion string, notificationID string,
...
-			if err != nil {
-				log.Println(err)
-				return err
-			}
 
+				err = infradb.client.Set("vpns", &vpns)
+				if err != nil {
+					log.Println(err)
+					return err
+				}
```
artek-koltun, line 996: This chunk looks very similar to the one above...
Can we create a dedicated function to do that and eliminate the duplication? Maybe `removeVniFromVpns`?

pkg/infradb/infradb.go
atulpatel261194: done

pkg/infradb/infradb.go
atulpatel261194: Done

pkg/infradb/infradb.go
```diff
@@ -257,6 +257,29 @@ func UpdateLB(lb *LogicalBridge) error {
...
+			log.Println("removeVniFromVpns(): No VPNs have been found")
+			log.Println(vni, ":")
+			return ErrKeyNotFound
+		}
+		delete(vpns, vni)
+
+		err = infradb.client.Set("vpns", &vpns)
+		if err != nil {
+			log.Println(err)
+			return err
```
artek-koltun, line 278: again, here we print an error msg and return that err.
On the level above, we print the same error. What value do we add by behaving that way? Probably we just flood the log with the identical error messages.
Each error should be handled once, and logging is a form of handling. You should either log it or return it, not both

artek-koltun (approved):

PR #386, opened 2024-05-20
feat(evpn-bridge): intel e2000 vendor plugin

atulpatel261194:


glimchb (changes requested):
I wonder is this is better to implement vendor plugin in their own repos and then import them here and use. we can't really maintain/code review that many plugins...

mardim91:
> I wonder is this is better to implement vendor plugin in their own repos and then import them here and use. we can't really maintain/code review that many plugins...

Hello @glimchb,

This is something that is our intention too. So we plan to do this in the future. We thought that is basic easier for now that we are implementing the whole rearchitecture thing and we test everything to put everything in one repo. When the main work is done then we can create different repos for the vendor plugins and inject them from there but for now let's keep everyhting in one place because that helps on the testing and doesn't introduce any extra integration effort.

I can open an issue to track this if you want

mardim91:
Hello @glimchb 

I have noticed that you have putted an "invalid" tag on this PR. What does that exactly mean? can you please explain ?

PR #387, opened 2024-05-22
fix(evpn): build fix

atulpatel261194:
pointer check is enough for deleting the vni from vpn's.


atulpatel261194:
Not required, as ping 6.6.6.6 worked

PR #398, opened 2024-09-11
Introduce replaydb feature

mardim91:
In order to move the evpn-gw-bridge into a more resilient path we have implemented the replaydb feature.

When a module (e.g. FRR module) confronts a permanent error then stops retrying and the replay procedure kicks in. The replay procedure will clean up all the system configuration (e.g. FRR daemon) that is related to the failed module and will replay the relevant objects from the database in order to bring the system configuration back on track again.

The replay feature currently is only supported for the FRR module and the FRR daemon. Support for all the modules will be added in the future

### DIFF
--- a/pkg/frr/frr.go
+++ b/pkg/frr/frr.go
@@ -217,7 +217,7 @@ func handlesvi(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
+		comp.CheckReplayThreshold(replayThreshold)
 
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -240,7 +240,7 @@ func handlesvi(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
+		comp.CheckReplayThreshold(replayThreshold)
 
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -318,7 +318,7 @@ func handlevrf(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
+		comp.CheckReplayThreshold(replayThreshold)
 
 		err := infradb.UpdateVrfStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -341,7 +341,7 @@ func handlevrf(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
+		comp.CheckReplayThreshold(replayThreshold)
 
 		err := infradb.UpdateVrfStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
--- a/pkg/frr/frr.go
+++ b/pkg/frr/frr.go
@@ -200,9 +200,7 @@ func handlesvi(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		if comp.Timer > replayThreshold {
-			comp.Replay = true
-		}
+		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
 
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -225,9 +223,7 @@ func handlesvi(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		if comp.Timer > replayThreshold {
-			comp.Replay = true
-		}
+		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
 
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -305,9 +301,7 @@ func handlevrf(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		if comp.Timer > replayThreshold {
-			comp.Replay = true
-		}
+		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
 
 		err := infradb.UpdateVrfStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
@@ -330,9 +324,7 @@ func handlevrf(objectData *eventbus.ObjectData) {
 		log.Printf("%+v\n", comp)
 
 		// Checking the timer to decide if we need to replay or not
-		if comp.Timer > replayThreshold {
-			comp.Replay = true
-		}
+		comp.Replay = utils.CheckReplayThreshold(comp.Timer, replayThreshold)
 
 		err := infradb.UpdateVrfStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
--- a/pkg/frr/frr.go
+++ b/pkg/frr/frr.go
@@ -32,22 +32,39 @@ import (
 // frrComp string constant
 const frrComp string = "frr"
 
-// runningFrrConfFile holds the running configuration of FRR daemon
-const runningFrrConfFile = "/etc/frr/frr.conf"
-
-// basicFrrConfFile holds the basic/initial configuration of FRR daemon
-const basicFrrConfFile = "/etc/frr/frr-basic.conf"
-
-// backupFrrConfFile holds the backup configuration the current running config of FRR daemon
-const backupFrrConfFile = "/etc/frr/frr.conf.bak"
-
+// replayThreshold time threshold for replay
 const replayThreshold = 64 * time.Second
 
 // ModulefrrHandler empty structure
 type ModulefrrHandler struct{}
 
 // ModuleFrrActionHandler empty structure
-type ModuleFrrActionHandler struct{}
+type ModuleFrrActionHandler struct {
+	// runningFrrConfFile holds the running configuration of FRR daemon
+	runningFrrConfFile string
+	// basicFrrConfFile holds the basic/initial configuration of FRR daemon
+	basicFrrConfFile string
+	// backupFrrConfFile holds the backup configuration the current running config of FRR daemon
+	backupFrrConfFile string
+}
+
+// NewModuleFrrActionHandler initializes a default ModuleFrrActionHandler
+func NewModuleFrrActionHandler() *ModuleFrrActionHandler {
+	return &ModuleFrrActionHandler{
+		runningFrrConfFile: "/etc/frr/frr.conf",
+		basicFrrConfFile:   "/etc/frr/frr-basic.conf",
+		backupFrrConfFile:  "/etc/frr/frr.conf.bak",
+	}
+}
+
+// NewModuleFrrActionHandlerWithArgs initializes a ModuleFrrActionHandler
+func NewModuleFrrActionHandlerWithArgs(runningFrrConfFile, basicFrrConfFile, backupFrrConfFile string) *ModuleFrrActionHandler {
+	return &ModuleFrrActionHandler{
+		runningFrrConfFile: runningFrrConfFile,
+		basicFrrConfFile:   basicFrrConfFile,
+		backupFrrConfFile:  backupFrrConfFile,
+	}
+}
 
 // HandleEvent handles the events
 func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.ObjectData) {
@@ -68,13 +85,13 @@ func (h *ModuleFrrActionHandler) HandleAction(actionType string, actionData *act
 	switch actionType {
 	case "preReplay":
 		log.Printf("Module FRR received %s\n", actionType)
-		handlePreReplay(actionData)
+		h.handlePreReplay(actionData)
 	default:
 		log.Printf("error: Unknown action type %s", actionType)
 	}
 }
 
-func handlePreReplay(actionData *actionbus.ActionData) {
+func (h *ModuleFrrActionHandler) handlePreReplay(actionData *actionbus.ActionData) {
 	var deferErr error
 
 	defer func() {
@@ -84,22 +101,22 @@ func handlePreReplay(actionData *actionbus.ActionData) {
 	}()
 
 	// Backup the current running config
-	deferErr = os.Rename(runningFrrConfFile, backupFrrConfFile)
+	deferErr = os.Rename(h.runningFrrConfFile, h.backupFrrConfFile)
 	if deferErr != nil {
 		log.Printf("FRR: handlePreReplay(): Failed to backup running config of FRR: %s\n", deferErr)
 		return
 	}
 
 	// Create a new running config based on the basic/initial FRR config
-	input, deferErr := os.ReadFile(basicFrrConfFile)
+	input, deferErr := os.ReadFile(h.basicFrrConfFile)
 	if deferErr != nil {
-		log.Printf("FRR: handlePreReplay(): Failed to read content of %s: %s\n", basicFrrConfFile, deferErr)
+		log.Printf("FRR: handlePreReplay(): Failed to read content of %s: %s\n", h.basicFrrConfFile, deferErr)
 		return
 	}
 
-	deferErr = os.WriteFile(runningFrrConfFile, input, 0600)
+	deferErr = os.WriteFile(h.runningFrrConfFile, input, 0600)
 	if deferErr != nil {
-		log.Printf("FRR: handlePreReplay(): Failed to write content to %s: %s\n", runningFrrConfFile, deferErr)
+		log.Printf("FRR: handlePreReplay(): Failed to write content to %s: %s\n", h.runningFrrConfFile, deferErr)
 		return
 	}
 
@@ -122,9 +139,9 @@ func handlePreReplay(actionData *actionbus.ActionData) {
 		return
 	}
 
-	deferErr = os.Chown(runningFrrConfFile, uid, gid)
+	deferErr = os.Chown(h.runningFrrConfFile, uid, gid)
 	if deferErr != nil {
-		log.Printf("FRR: handlePreReplay(): Failed to chown of %s to frr:frr : %s\n", runningFrrConfFile, deferErr)
+		log.Printf("FRR: handlePreReplay(): Failed to chown of %s to frr:frr : %s\n", h.runningFrrConfFile, deferErr)
 		return
 	}
 
@@ -368,7 +385,7 @@ func subscribeInfradb(config *config.Config) {
 			}
 		}
 	}
-	ab.StartSubscriber(frrComp, "preReplay", &ModuleFrrActionHandler{})
+	ab.StartSubscriber(frrComp, "preReplay", NewModuleFrrActionHandler())
 }
 
 // ctx variable of type context
--- a/pkg/frr/frr.go
+++ b/pkg/frr/frr.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"os/user"
 	"path"
 	"reflect"
 	"strconv"
@@ -23,6 +24,7 @@ import (
 	"github.com/opiproject/opi-evpn-bridge/pkg/config"
 	"github.com/opiproject/opi-evpn-bridge/pkg/infradb"
 	"github.com/opiproject/opi-evpn-bridge/pkg/infradb/common"
+	"github.com/opiproject/opi-evpn-bridge/pkg/infradb/subscriberframework/actionbus"
 	"github.com/opiproject/opi-evpn-bridge/pkg/infradb/subscriberframework/eventbus"
 	"github.com/opiproject/opi-evpn-bridge/pkg/utils"
 )
@@ -30,8 +32,22 @@ import (
 // frrComp string constant
 const frrComp string = "frr"
 
+// runningFrrConfFile holds the running configuration of FRR daemon
+const runningFrrConfFile = "/etc/frr/frr.conf"
+
+// basicFrrConfFile holds the basic/initial configuration of FRR daemon
+const basicFrrConfFile = "/etc/frr/frr-basic.conf"
+
+// backupFrrConfFile holds the backup configuration the current running config of FRR daemon
+const backupFrrConfFile = "/etc/frr/frr.conf.bak"
+
+const replayThreshold = 64 * time.Second
+
 // ModulefrrHandler empty structure
 type ModulefrrHandler struct{}
+
+// ModuleFrrActionHandler empty structure
+type ModuleFrrActionHandler struct{}
 
 // HandleEvent handles the events
 func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.ObjectData) {
@@ -45,6 +61,76 @@ func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.Ob
 	default:
 		log.Printf("error: Unknown event type %s", eventType)
 	}
+}
+
+// HandleAction handles the actions
+func (h *ModuleFrrActionHandler) HandleAction(actionType string, actionData *actionbus.ActionData) {
+	switch actionType {
+	case "preReplay":
+		log.Printf("Module FRR received %s\n", actionType)
+		handlePreReplay(actionData)
+	default:
+		log.Printf("error: Unknown action type %s", actionType)
+	}
+}
+
+func handlePreReplay(actionData *actionbus.ActionData) {
+	var deferErr error
+
+	defer func() {
+		// The ErrCh is used in order to notify the sender that the preReplay step has
+		// been executed successfully.
+		actionData.ErrCh <- deferErr
+	}()
+
+	// Backup the current running config
+	if err := os.Rename(runningFrrConfFile, backupFrrConfFile); err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to backup running config of FRR: %s\n", err)
+		deferErr = err
+		return
+	}
+
+	// Create a new running config based on the basic/initial FRR config
+	input, err := os.ReadFile(basicFrrConfFile)
+	if err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to read content of %s: %s\n", basicFrrConfFile, err)
+		deferErr = err
+		return
+	}
+
+	if err := os.WriteFile(runningFrrConfFile, input, 0600); err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to write content to %s: %s\n", runningFrrConfFile, err)
+		deferErr = err
+		return
+	}
+
+	// Change ownership of the frr.conf to frr:frr
+	group, err := user.Lookup("frr")
+	if err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to lookup user frr %s\n", err)
+		deferErr = err
+		return
+	}
+
+	uid, _ := strconv.Atoi(group.Uid)
+	gid, _ := strconv.Atoi(group.Gid)
+
+	if err := os.Chown(runningFrrConfFile, uid, gid); err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to chown of %s to frr:frr : %s\n", runningFrrConfFile, err)
+		deferErr = err
+		return
+	}
+
+	// Restart FRR daemon
+	_, errCmd := utils.Run([]string{"systemctl", "restart", "frr"}, false)
+	if errCmd != 0 {
+		log.Println("FRR: handlePreReplay(): Failed to restart FRR daemon")
+		err := fmt.Errorf("restart FRR daemon failed")
+		deferErr = err
+		return
+	}
+
+	log.Println("FRR: handlePreReplay(): The pre-replay procedure has executed successfully")
 }
 
 // handlesvi handles the svi functionality
@@ -106,6 +192,12 @@ func handlesvi(objectData *eventbus.ObjectData) {
 			comp.CompStatus = common.ComponentStatusError
 		}
 		log.Printf("%+v\n", comp)
+
+		// Checking the timer to decide if we need to replay or not
+		if comp.Timer > replayThreshold {
+			comp.Replay = true
+		}
+
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
 			log.Printf("error in updating svi status: %s\n", err)
@@ -125,6 +217,12 @@ func handlesvi(objectData *eventbus.ObjectData) {
 			comp.CompStatus = common.ComponentStatusError
 		}
 		log.Printf("%+v\n", comp)
+
+		// Checking the timer to decide if we need to replay or not
+		if comp.Timer > replayThreshold {
+			comp.Replay = true
+		}
+
 		err := infradb.UpdateSviStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
 			log.Printf("error in updating svi status: %s\n", err)
@@ -199,6 +297,12 @@ func handlevrf(objectData *eventbus.ObjectData) {
 			comp.CompStatus = common.ComponentStatusError
 		}
 		log.Printf("%+v\n", comp)
+
+		// Checking the timer to decide if we need to replay or not
+		if comp.Timer > replayThreshold {
+			comp.Replay = true
+		}
+
 		err := infradb.UpdateVrfStatus(objectData.Name, objectData.ResourceVersion, objectData.NotificationID, nil, comp)
 		if err != nil {
 			log.Printf("error in updating vrf status: %s\n", err)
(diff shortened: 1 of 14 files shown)


pkg/infradb/infradb.go
```diff
@@ -1340,28 +1334,27 @@ func UpdateSviStatus(name string, resourceVersion string, notificationID string,
...
-		if comp.Name == component.Name {
-			svi.Status.Components[i] = component
+	if component.Replay {
+		// One of the components has requested a replay of the DB.
+		// The task related to the status update will be dropped.
+		log.Printf("UpdateSviStatus(): Component %s has requested a replay\n", component.Name)
+		taskmanager.TaskMan.StatusUpdated(svi.Name, "svi", svi.ResourceVersion, notificationID, true, &component)
+		go startReplayProcedure(component.Name)
+		return nil
+	}
```
artek-koltun, line 1344: the same about this block of code. It looks similar for each type

pkg/infradb/subscriberframework/actionbus/actionbus.go
```diff
@@ -0,0 +1,136 @@
...
+	defer a.subscriberL.Unlock()
+
+	subscriber := &Subscriber{
+		Name: moduleName,
+		Ch:   make(chan interface{}),
+		Quit: make(chan bool),
+	}
+
+	a.subscribers[actionType] = append(a.subscribers[actionType], subscriber)
+	a.actionHandlers[moduleName+"."+actionType] = actionHandler
```
artek-koltun, line 99: if we rely on string concatenation somewhere, we need to make sure that the concatenation is placed into a  dedicated function and used where it is needed. It is too easy to make a typo and get a mistake

pkg/infradb/subscriberframework/actionbus/actionbus.go
```diff
@@ -0,0 +1,136 @@
...
+	log.Printf("Subscriber %s registered for action %s\n", moduleName, actionType)
+	return subscriber
+}
+
+// GetSubscribers api is used to fetch the list of subscribers registered with given actionType
+func (a *ActionBus) GetSubscribers(actionType string) []*Subscriber {
+	a.mutex.RLock()
+	defer a.mutex.RUnlock()
+
+	return a.subscribers[actionType]
```
artek-koltun, line 110: if we expose a slice of pointers to another thread here, how can we be sure that no one would change that slice from another thread? not change elements from another thread?

pkg/infradb/subscriberframework/actionbus/actionbus.go
```diff
@@ -0,0 +1,136 @@
...
+	a.actionHandlers[moduleName+"."+actionType] = actionHandler
+
+	log.Printf("Subscriber %s registered for action %s\n", moduleName, actionType)
+	return subscriber
+}
+
+// GetSubscribers api is used to fetch the list of subscribers registered with given actionType
+func (a *ActionBus) GetSubscribers(actionType string) []*Subscriber {
+	a.mutex.RLock()
+	defer a.mutex.RUnlock()
```
artek-koltun, line 108: different mutexes are used to access `subscribers` this does not look correct

pkg/infradb/taskmanager/taskmanager.go
```diff
@@ -103,7 +105,15 @@ func (t *TaskManager) StatusUpdated(name, objectType, resourceVersion, notificat
...
 
+// ReplayFinished notifies that the replay of objects has finished
+func (t *TaskManager) ReplayFinished() {
+	close(t.replayChan)
+	log.Println("ReplayFinished(): Replay has finished.")
+}
+
 // processTasks processes the task
+//
+//gocognit:ignore
```
artek-koltun, line 116: I suppose it won't take much time to create a dedicated function rather than suppress warnings

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
+// SPDX-License-Identifier: Apache-2.0
```
artek-koltun, line 1: apparently, the file should be named replay.go

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+		if preSub.Name == componentName {
+			preSubscriber = preSub
+			break
+		}
+	}
+
+	if preSubscriber == nil {
+		deferErr = fmt.Errorf("no pre-replay subscriber for %s", componentName)
+		log.Printf("startReplayProcedure(): Error %+v\n", deferErr)
+		return
```
artek-koltun, line 43: global lock  won't be released if we exit here, right?

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+			break
+		}
+	}
+
+	for _, bpSub := range bpSubs {
+		if bpSub.Name == componentName {
+			objectTypesToReplay = append(objectTypesToReplay, "bridge-port")
+			break
+		}
+	}
```
artek-koltun, line 121: the same operation is repeated 4 times, apparently there is a way to have only one for here

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+	for _, sviSub := range sviSubs {
+		if sviSub.Name == componentName {
+			objectTypesToReplay = append(objectTypesToReplay, "svi")
+			break
+		}
+	}
+
+	for _, bpSub := range bpSubs {
+		if bpSub.Name == componentName {
+			objectTypesToReplay = append(objectTypesToReplay, "bridge-port")
```
artek-koltun, line 118: why bare strings are used for this purpose. Strings are dangerous since we can easily make a typo. It is much more reliable to use enums(int or string)

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+			break
+		}
+	}
+
+	return objectTypesToReplay
+}
+
+// nolint: funlen, gocognit
```
artek-koltun, line 126: I'd prefer to fix linter warnings, rather than suppressing them.

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+func gatherObjectsAndSubsToReplay(componentName string, objectTypesToReplay []string) ([]interface{}, [][]*eventbus.Subscriber, error) {
+	objectsToReplay := []interface{}{}
+	subsForReplay := [][]*eventbus.Subscriber{}
+
+	bpSubs := eventbus.EBus.GetSubscribers("bridge-port")
+	sviSubs := eventbus.EBus.GetSubscribers("svi")
+	lbSubs := eventbus.EBus.GetSubscribers("logical-bridge")
+	vrfSubs := eventbus.EBus.GetSubscribers("vrf")
```
artek-koltun, line 134: can we be sure that number of subscribers won't change in between getObjectTypesToReplay gatherObjectsAndSubsToReplay calls?

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+				// for replay
+				tempSubs := bp.prepareObjectsForReplay(componentName, bpSubs)
+
+				err = infradb.client.Set(bp.Name, bp)
+				if err != nil {
+					return nil, nil, err
+				}
+				subsForReplay = append(subsForReplay, tempSubs)
+				objectsToReplay = append(objectsToReplay, bp)
+			}
```
artek-koltun, line 273: again, the code in all cases has similarity. Can we unify it instead of just duplicating?

pkg/infradb/utils.go
```diff
@@ -0,0 +1,294 @@
...
+		switch tempObj := obj.(type) {
+		case *Vrf:
+			taskmanager.TaskMan.CreateTask(tempObj.Name, "vrf", tempObj.ResourceVersion, subsForReplay[i])
+		case *LogicalBridge:
+			taskmanager.TaskMan.CreateTask(tempObj.Name, "logical-bridge", tempObj.ResourceVersion, subsForReplay[i])
+		case *Svi:
+			taskmanager.TaskMan.CreateTask(tempObj.Name, "svi", tempObj.ResourceVersion, subsForReplay[i])
+		case *BridgePort:
+			taskmanager.TaskMan.CreateTask(tempObj.Name, "bridge-port", tempObj.ResourceVersion, subsForReplay[i])
+		}
```
artek-koltun, line 292: what if we receive type not listed in this switch?

pkg/utils/helpers.go
```diff
@@ -22,6 +23,22 @@ func ApplyMaskToStoredPbObject[T proto.Message](updateMask *fieldmaskpb.FieldMas
 	fieldmask.Update(updateMask, dst, src)
 }
 
+// Run function run the commands
+func Run(cmd []string, _ bool) (string, int) {
+	var out []byte
+	var err error
+	out, err = exec.Command(cmd[0], cmd[1:]...).CombinedOutput() //nolint:gosec
+	if err != nil {
+		/*if flag {
```
artek-koltun, line 32: if we don't need code, it can be just removed

pkg/frr/frr.go
```diff
@@ -47,6 +63,76 @@ func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.Ob
...
+	switch actionType {
+	case "preReplay":
+		log.Printf("Module FRR received %s\n", actionType)
+		handlePreReplay(actionData)
+	default:
+		log.Printf("error: Unknown action type %s", actionType)
+	}
+}
+
+func handlePreReplay(actionData *actionbus.ActionData) {
```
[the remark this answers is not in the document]
mardim91: Sorry didn't get you comment. Which consts are you talking about ?

pkg/frr/frr.go
```diff
@@ -47,6 +63,76 @@ func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.Ob
...
+	var deferErr error
+
+	defer func() {
+		// The ErrCh is used in order to notify the sender that the preReplay step has
+		// been executed successfully.
+		actionData.ErrCh <- deferErr
+	}()
+
+	// Backup the current running config
+	if err := os.Rename(runningFrrConfFile, backupFrrConfFile); err != nil {
```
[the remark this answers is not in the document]
mardim91: yes maybe I can do that. I thought will be just cleaner from readability perspective but if you think is better to directly assign it to deferErr then I can do that.

pkg/frr/frr.go
```diff
@@ -47,6 +63,76 @@ func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.Ob
...
+	// Change ownership of the frr.conf to frr:frr
+	group, err := user.Lookup("frr")
+	if err != nil {
+		log.Printf("FRR: handlePreReplay(): Failed to lookup user frr %s\n", err)
+		deferErr = err
+		return
+	}
+
+	uid, _ := strconv.Atoi(group.Uid)
+	gid, _ := strconv.Atoi(group.Gid)
```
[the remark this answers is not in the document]
mardim91: Let me check that. Maybe I have just missed it

pkg/frr/frr.go
```diff
@@ -47,6 +63,76 @@ func (h *ModulefrrHandler) HandleEvent(eventType string, objectData *eventbus.Ob
...
+	_, errCmd := utils.Run([]string{"systemctl", "restart", "frr"}, false)
+	if errCmd != 0 {
+		log.Println("FRR: handlePreReplay(): Failed to restart FRR daemon")
+		err := fmt.Errorf("restart FRR daemon failed")
+		deferErr = err
+		return
+	}
+
+	log.Println("FRR: handlePreReplay(): The pre-replay procedure has executed successfully")
+}
```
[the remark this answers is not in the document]
mardim91: The whole reason that we do the replay procedure is because the frr configuration is malformed for whatever reason and we want to repalce it with a clean base configuration and replay the db obects in order to reapply the dynamic configuration. So there is no point to restore the previous configuration in case of error as this malformed configuration was the reason that the replay procedure started in first place.

pkg/frr/frr.go
```diff
@@ -218,6 +322,12 @@ func handlevrf(objectData *eventbus.ObjectData) {
 			comp.CompStatus = common.ComponentStatusError
 		}
 		log.Printf("%+v\n", comp)
+
+		// Checking the timer to decide if we need to replay or not
+		if comp.Timer > replayThreshold {
```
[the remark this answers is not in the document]
mardim91: what exactly you mean by saying "only one check" ? Like creating a common function and calling it from both device types related functions ?

pkg/infradb/bridge.go
```diff
@@ -211,3 +211,54 @@ func (in *LogicalBridge) DeleteBridgePort(bpName, bpMac string) error {
...
+			tempSubs = append(tempSubs, lbSubs[i])
+		}
+	}
+	if in.Status.LBOperStatus == LogicalBridgeOperStatusUp {
+		in.Status.LBOperStatus = LogicalBridgeOperStatusDown
+	}
+
+	in.ResourceVersion = generateVersion()
+	return tempSubs
+}
```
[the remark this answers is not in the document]
mardim91: I have spent some thoughts on this myself in the past. I haven't concluded myself if that will be better or will increase the difficulty of the readability of the code. The way that is now helps us to debug any potential issues better as we have heavy functional deliveries until the end of the year.  I would prefer to leave it as it is for now that we are more focused on the functional deliveries. We can revisit this topic in the future that we will have more time and we can include it in the general restructuring topic. I can open an issue for this to track it if you want

pkg/infradb/infradb.go
mardim91: Same answer as above.

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Do you think it worths to create a new function for just one line of code. Also the concatenation is based on variables not text (except the "." of course) so the golang compiler will obect during the build if there is any typo. WDYT ?

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Sorry didn't understand what you mean here. Can you please elaborate a bit ?

pkg/infradb/taskmanager/taskmanager.go
mardim91: Let me see what I can do for this

pkg/infradb/utils.go
mardim91: Maybe. I will chck it out

pkg/infradb/utils.go
mardim91: I think you are right on this. I will investigate

mardim91: I think you are right on this.  I will investigate

pkg/infradb/utils.go
mardim91: I think you are right on this. We can include that in the whole restructuring effor into the future. I will open an issue to track this one because we have this problem I believe in several places into the code. I hope that is allright

pkg/infradb/utils.go
mardim91: I will see what I can do.

pkg/infradb/utils.go
mardim91: Is not easy to make this function smaller. It will create confusion and it will become error prone. I think I will ass this in the context of the restructuring effort. I can open an issue for that.

pkg/infradb/utils.go
mardim91: Yes I have tried that and I didn't see any obvious solution so the code doesn't become error prone and unreadable. I will open an issue for this one to revisit in the restructuring effort in the future. Hope that is alright

pkg/infradb/utils.go
mardim91: The types of objects are just 4. Always. I do not see any case where we will receive any other type.

pkg/utils/helpers.go
mardim91: I will check on this

mardim91 (commented):
Comments addresed

pkg/frr/frr.go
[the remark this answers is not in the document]
mardim91: Done

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Sorry I am not sure that I get this one. Can you please elaborate ? For which reason someone to change the slice of pointers ?

pkg/infradb/taskmanager/taskmanager.go
mardim91: Done

pkg/infradb/utils.go
mardim91: Done

pkg/infradb/utils.go
mardim91: Done

pkg/utils/helpers.go
mardim91: Done

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: The const in the file we have `ModuleFrrActionHandler`
```go
// runningFrrConfFile holds the running configuration of FRR daemon
const runningFrrConfFile = "/etc/frr/frr.conf"

// basicFrrConfFile holds the basic/initial configuration of FRR daemon
const basicFrrConfFile = "/etc/frr/frr-basic.conf"

// backupFrrConfFile holds the backup configuration the current running config of FRR daemon
const backupFrrConfFile = "/etc/frr/frr.conf.bak"

const replayThreshold = 64 * time.Second
```

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: yes

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: I mean if we are going to restore renamed configuration files and etc.

pkg/infradb/subscriberframework/actionbus/actionbus.go
artek-koltun: here we return `a.subscribers[actionType]` from this function. It is a slice. Slices are implemented in such a way that they copy just a pointer to actual data. It means that whoever called that function can change subscribers stored in `a.subscribers` (if they assumed that they keep a deep copy of it)

pkg/infradb/subscriberframework/actionbus/actionbus.go
artek-koltun: you have a resource `a.subscribers` If you protect a resource with a mutex, it means that you need to lock the same mutex each time you access the resource. Here I see, you use 2 different mutexes to access the same resource `a.mutex` and `a.subscriberL`

artek-koltun: or since you use channels, I would think how we could avoid using mutexes at all by transferring ownership/making a deep copy/etc.

pkg/infradb/utils.go
artek-koltun: looks convoluted. We unlock in 2 different places. In one if there is an error, in another if there is no error.
I would move the code where we need to keep lock in a separate function where just `defer mu.Unlock()` is used

pkg/infradb/utils.go
artek-koltun: why? They look similar, you can move the code into the corresponding function/functions

pkg/infradb/utils.go
artek-koltun: do we have a way to prevent adding new subscribers when we think that all of them are created?

pkg/infradb/utils.go
artek-koltun: we can receive if we have a mistake -> why not handle such a case instead of crashing

pkg/infradb/bridge.go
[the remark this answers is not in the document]
artek-koltun: Each such duplication increases maintenance efforts and complexity. I do not see anyone is working on it and it annoys me

pkg/frr/frr.go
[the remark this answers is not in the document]
mardim91: No its not a method of the ModuleFrrActionHandler struct. It is just a separate function. We could I believe make it a method of the ModuleFrrActionHandler in order to inherit the constant variables from the struct ModuleFrrActionHandler. Do you think is it worth it ?

pkg/infradb/bridge.go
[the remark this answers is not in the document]
mardim91: I can fully understand your frustration. We also on our side have limited resources and is inevitable some tasks to have bigger priorities than other. The restructuring we always have in the back of our minds but we just didn't find time or resources to assign it to somebody. I would raise the issue internally to see what I can do about this. 

Also about those functions that you have mentioned above. Maybe they look similar but I will not say that it is duplication because they handle different things which are related to the object that they are dealing with. So from my perspective keeps the code cleaner. Yes maybe they have the same name but they are methods to different structures and their bodies I am not sure if will be made more confusing if we try to fit them in a common function. I can take a look on this and see if I can somehow make something better.

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: ok I will create a function then that does this concatenation

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Let me think about this

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: I need to think about this

pkg/infradb/utils.go
mardim91: I think the question here is "from whom do we expect to add new subscribers ?" During runtime I do not see any process adding new subscribers. I do not think there is any trap here that can cause problems. So who do we expect to add new subscribers ?

pkg/infradb/utils.go
mardim91: I can put a "default" case if that is necessary but I can not see how we will receive any other type except from those 4 types that we allready have.

pkg/infradb/bridge.go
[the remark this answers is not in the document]
mardim91: Hello @artek-koltun,

I have taken a look into prepareObjectsForReplay function. I really didn't find any sart way to avaoid having the function as a method of the different objects (VRF,LB,BP,SVI) can you please also check and tell me if you have something to suggest ? The way that is now seems ok to me but if you know any better way please let me know.

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: The whole implementation is based on pointers so we can have better utilization of memory and alo there are channels for each subscriber that they used for communication between the different threads. So I do not think is that easy to just return a copy of Subscribers for instance and not the pointers of those subscribers as then the system will not fucntion due to the channels that are bound to those subscribers objects. I hear your concern but I honestly cannot see any case where some thread will change a subscriber. Even if a thread can change a subscriber I do not see a real risk that this will happen. Yes is a valid concern but I see low possibility of happening. If we return a copy of the Subscribers and not pointers then the whole implementation I believe needs to change in many places and maybe we need to rethink the whole thing which is not vital right now I believe. Also the code is functioning as one we do not have plugins that come and go during runtime so if the code passes the integration tests and the tests in general then I do not see how some thread that have been introduced through some PR will change the subscribers and the integration test gate will give a green light.

pkg/infradb/utils.go
mardim91: I can probably create 4 functions that each one of those handle the different types of cases/objects (vrf,lb,bp,svi) in the switch statement. Is that ok ? I cann not create one function to handle all the cases as we are talking about different structs to be handled by the same function and that is really difficult to implement in golang except if you have something else to propose. But 4 functions one per case I can create without a problem. WDYT ?

pkg/frr/frr.go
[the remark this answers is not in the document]
mardim91: Done

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Done

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Done

pkg/infradb/utils.go
mardim91: I have changed it. I think now is better

pkg/infradb/utils.go
mardim91: Done

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: why not? For example, when you write unit tests, you can replace file names with test related names and check with real files (which is always more reliable than mocking)

pkg/infradb/subscriberframework/actionbus/actionbus.go
artek-koltun: but in `GetSubscribers` you still can use `RLock` with `subscriberL`, right?

pkg/infradb/subscriberframework/actionbus/actionbus.go
artek-koltun: Maybe `ComposeActionHandlerName` instead of `ConcatenateModuleNameWithType`?

pkg/infradb/utils.go
artek-koltun: pls create an issue/add to the existing one

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: yes you are right I missed that. I will fix it

pkg/infradb/utils.go
artek-koltun: Generics(less preferable)/interfaces(more preferable)?
If `infradb.client.` could return not only the concrete types, but also some common interface(slice of such interface instances) with required operations like `GetName`, `IsReplayNeeded`, `GetStatus` then I think it would be possible

pkg/infradb/bridge.go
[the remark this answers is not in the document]
artek-koltun: maybe some variation of [template method](https://refactoring.guru/design-patterns/template-method) help here?

pkg/infradb/subscriberframework/actionbus/actionbus.go
artek-koltun: `ComposeHandlerName`?

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: Would it make sense to move it entirely into comp?
```go
func CheckReplayThreshold(replayThreshold time.Duration) {
	replay = (timer > replayThreshold)
}
```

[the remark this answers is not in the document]
mardim91: Sorry didn't get this one. Where you think we should move the function ? Can you please elaboate a bit ?

pkg/frr/frr.go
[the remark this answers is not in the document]
mardim91: Done

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Done

pkg/infradb/subscriberframework/actionbus/actionbus.go
mardim91: Done

pkg/infradb/bridge.go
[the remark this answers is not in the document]
mardim91: I checked the template method that you pointed out above. 

This template method wroks better when you have native support for inheritance in the language (e.g. Java). It is a bit tricky and cumbersome to do it in golang and I am not 100% sure that will make the code simpler and more readbale.  Either way if we want to go down that path I believe that this needs to happen properly as a refactoring activity as requires changes in many places which exceed the limits of this feature PR. Also requires proper planning and thinking do be sure how we want to structure the code in order to make it more flexible and also not breaking anything. 

This is an activity that I do not think is suited for this PR and also can break the functionality easily. I think is more safe to do it later as refactoring activity so we have full control on what we are doing and aslo be sure that we do not break anything as we will have more time in our hands to do it properly. I can open an issue to put a hint on the template method that you have pointed out so in the future we know where to look.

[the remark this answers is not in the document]
mardim91: I checked the template method that you pointed out above. 

This template method works better when you have native support for inheritance in the language (e.g. Java). It is a bit tricky and cumbersome to do it in golang and I am not 100% sure that will make the code simpler and more readable. Either way if we want to go down that path I believe that this needs to happen properly as a refactoring activity as requires changes in many places which exceed the limits of this feature PR. Also requires proper planning and thinking to be sure how we want to structure the code in order to make it more flexible and also not breaking anything. 

This is an activity that I do not think is suited for this PR and also can break the functionality easily. I think is more safe to do it later as refactoring activity so we have full control on what we are doing and aslo be sure that we do not break anything as we will have more time in our hands to do it properly. I can open an issue and put a hint on the template method that you have pointed out so in the future we know where to look. 

Bottom line is that we need to have a proper design first and be sure that this template method works without any problem in golang before we go in and starting doing functional changes. So I think this is something for the future where we will have enough time to think about this and decide what changes we need to do in general in order to make the code better in many places.

pkg/infradb/utils.go
mardim91: Yes I am not sure how to change this right now.  I think we should think this through better when we have more time. Let's add this to the refactoring topic. I can open a github issue to track this if that is alright. I will suggest to park this for now until we have a good design that will address the refactorng topic in general

pkg/infradb/bridge.go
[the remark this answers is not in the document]
artek-koltun: > This template method works better when you have native support for inheritance in the language (e.g. Java).

Go embedding and interfaces are even more flexible than inheritance in Java and I believe that template method works well for Go as well

I think the commit you provided is only a midway of a rework. `setStatusComponents`, `getStatusComponents`, `isOperationalStatus`, `setOperationalStatus`, `setNewResourceVersion` look very similar in all types, and it means that we could move the impl into the `Domain` struct with related struct members needed to run those methods. And `Domain` will be embedded into vrf/svi etc. (in the language of OOP it is very close to "vrf inherits `Domain`"). Also, it would require to have a single enum for status for all types I think it won't be a big problem since I see in the switches only up, down, toBeDeleted and unspecified. Are those enums used in proto representation?
Basically, templated method is not needed here, we can proceed with bare embedding.

It is true that rework can be broader than the scope of this PR. What I want you to do is to branch off here and create a draft PR where we could start to experiment. The commit here can be deleted.

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: Is it possible to move `CheckReplayThreshold` entirely into `comp` type? Then we don't need to pass the same `comp.Timer` all the time as well as assign the result to the same `comp.Replay`. That `comp` just have a method which handles all of that and work with its own members. Wondering if `replayThreshold` as well as this chunk of code
```go
               if status {
			comp.CompStatus = common.ComponentStatusSuccess
			comp.Timer = 0
		} else {
			if comp.Timer == 0 {
				comp.Timer = 2 * time.Second
			} else {
				comp.Timer *= 2
			}
			comp.CompStatus = common.ComponentStatusError
		}
```
can also go into `comp`?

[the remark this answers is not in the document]
mardim91: I think we can do many impovements here and we can do them in the experiment branch that I have created. The reason for this is:
1. This code snippet that you have posted above is common in all the modules and not just FRR. So changes in that code will touch all the components and exceeds the scope of this PR so let's do it better on the experiment branch to do it properly
2. The replayThreshold variable is decided by each module itself and that is why we cannot pass it to the comp structure that easily. If we want to pass it in that Comp structure then we need to find a clever way to do so. That means that it needs a bit more thinking and changes that will affect other part of the code that are outside of this scope. So let's do it to the experiment branch

The rest of your comments I have fixed

pkg/infradb/bridge.go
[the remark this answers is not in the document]
mardim91: yes the Enums are in the proto definition. I think we could change them and use one generic Enum. I need to think about this

https://github.com/opiproject/opi-api/blob/main/network/evpn-gw/l3_xpu_infra_mgr.proto#L338

I have created an experimentation branch: https://github.com/mardim91/opi-evpn-bridge/tree/experiment-refactoring

we can use this one for the refactoring attempts

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: for me it sounds strange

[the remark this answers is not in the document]
mardim91: I mean what is the point of restoring the renamed configuration files in case of error ? Those old configuration files that has been renamed are malformed (this is the reason that we are doing the replay procedure). So even if we restore them in case of some error we will still have malformed files so not sure what we gain with that.

pkg/frr/frr.go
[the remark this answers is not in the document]
artek-koltun: > This code snippet that you have posted above is common in all the modules and not just FRR. So changes in that code will touch all the components and exceeds the scope of this PR so let's do it better on the experiment branch to do it properly

You just added it in this PR, why do you need to rework the rest?

> The replayThreshold variable is decided by each module itself and that is why we cannot pass it to the comp structure that easily.

It is the corresponding handle*() call which creates com, and can pass the threshold at construction?

I just do not want we continue adding debt. If it does not require to rework other components, so I'd prefer it is fixed in place
This change looks pretty straighforward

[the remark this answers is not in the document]
mardim91: > You just added it in this PR, why do you need to rework the rest?

I didnt added this code snippet to this PR that was allready there.

```if status {
			comp.CompStatus = common.ComponentStatusSuccess
			comp.Timer = 0
		} else {
			if comp.Timer == 0 {
				comp.Timer = 2 * time.Second
			} else {
				comp.Timer *= 2
			}
			comp.CompStatus = common.ComponentStatusError
		}


```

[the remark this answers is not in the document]
mardim91: Are you talking about this comp here ? https://github.com/opiproject/opi-evpn-bridge/blob/c9baa3ee74869d690c7c63751572252165071946/pkg/frr/frr.go#L163

This is not the same comp object which we use in order to call the CheckReplayThreshold function

The object that we use to call this function is created here: https://github.com/opiproject/opi-evpn-bridge/blob/c9baa3ee74869d690c7c63751572252165071946/pkg/frr/frr.go#L199 

an is a different object so even if we add the replay threshold to the object in the line 163 will not be there in the line 199 because we call a function with a different object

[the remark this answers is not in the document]
mardim91: That's why I am saying that we need to find a better smarter way to handle this in the refactoring. Also the modules in general have duplicate code that can be written better so let's handle this also in the refactoring. WDYT ?

pkg/frr/frr.go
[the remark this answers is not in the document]
mardim91: Let's assume that we restore the old configuration in case of error. How do you suggest to do it ?

pkg/infradb/bridge.go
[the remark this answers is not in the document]
artek-koltun: pls note it and other topics somewhere in the issue, or in that dedicated PR

artek-koltun (approved):
The debt is growing

mardim91:
https://github.com/opiproject/opi-evpn-bridge/issues/405

PR #400, opened 2024-09-23
feat(idpool): idpool feature for generating id's

venkyvsp:
idpool feature for generating id's for mod pointer, routing table id , trie pointer.

### DIFF
--- a/pkg/LinuxGeneralModule/lgm.go
+++ b/pkg/LinuxGeneralModule/lgm.go
@@ -462,7 +462,6 @@ func setUpBridge(lb *infradb.LogicalBridge) (string, bool) {
 //nolint:funlen,gocognit
 func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	IPMtu := fmt.Sprintf("%+v", ipMtu)
-	var addKey int
 	if path.Base(vrf.Name) == "GRD" {
 		vrf.Metadata.RoutingTable = make([]*uint32, 2)
 		vrf.Metadata.RoutingTable[0] = new(uint32)
@@ -475,21 +474,15 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	vrf.Metadata.RoutingTable[0] = new(uint32)
 	var routingtable uint32
 	name := vrf.Name
-	for {
-		routingtable, _ = RouteTableGen.GetID(name, 0)
-		log.Printf("LGM assigned id %+v for vrf name %s\n", routingtable, vrf.Name)
-		isbusy, err := routingtableBusy(routingtable)
-		if err != nil {
-			log.Printf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingtable, err)
-			return fmt.Sprintf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingtable, err), false
-		}
-		if !isbusy {
-			log.Printf("LGM: Routing Table %d is not busy\n", routingtable)
-			break
-		}
-		log.Printf("LGM: Routing Table %d is busy\n", routingtable)
-		addKey++
-		name = fmt.Sprintf("%s%d", name, addKey)
+	routingtable, _ = RouteTableGen.GetID(name, 0)
+	log.Printf("LGM assigned id %+v for vrf name %s\n", routingtable, vrf.Name)
+	isbusy, err := routingtableBusy(routingtable)
+	if err != nil {
+		log.Printf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingtable, err)
+		return "", false
+	}
+	if !isbusy {
+		log.Printf("LGM: Routing Table %d is not busy\n", routingtable)
 	}
 	var vtip string
 	if !reflect.ValueOf(vrf.Spec.VtepIP).IsZero() {
--- a/pkg/LinuxGeneralModule/lgm.go
+++ b/pkg/LinuxGeneralModule/lgm.go
@@ -360,6 +360,7 @@ var RouteTableGen utils.IDPool
 // Initialize initializes the config, logger and subscribers
 func Initialize() {
 	eb := eventbus.EBus
+	var ok bool
 	for _, subscriberConfig := range config.GlobalConfig.Subscribers {
 		if subscriberConfig.Name == lgmComp {
 			for _, eventType := range subscriberConfig.Events {
@@ -370,8 +371,7 @@ func Initialize() {
 	brTenant = "br-tenant"
 	ipMtu = config.GlobalConfig.LinuxFrr.IPMtu
 	ctx = context.Background()
-	RouteTableGen = utils.IDPoolInit("RTtable", routingTableMin, routingTableMax)
-	if reflect.DeepEqual(RouteTableGen, utils.IDPool{}) {
+	if RouteTableGen, ok = utils.IDPoolInit("RTtable", routingTableMin, routingTableMax); !ok {
 		log.Printf("LGM: Failed in the assigning id \n")
 		return
 	}
--- a/pkg/LinuxGeneralModule/lgm.go
+++ b/pkg/LinuxGeneralModule/lgm.go
@@ -371,6 +371,10 @@ func Initialize() {
 	ipMtu = config.GlobalConfig.LinuxFrr.IPMtu
 	ctx = context.Background()
 	RouteTableGen = utils.IDPoolInit("RTtable", routingTableMin, routingTableMax)
+	if reflect.DeepEqual(RouteTableGen, utils.IDPool{}) {
+		log.Printf("LGM: Failed in the assigning id \n")
+		return
+	}
 	nlink = utils.NewNetlinkWrapperWithArgs(false)
 	// Set up the static configuration parts
 	_, err := nlink.LinkByName(ctx, brTenant)
--- a/pkg/LinuxGeneralModule/lgm.go
+++ b/pkg/LinuxGeneralModule/lgm.go
@@ -478,7 +478,7 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	vrf.Metadata.RoutingTable[0] = new(uint32)
 	var routingtable uint32
 	name := vrf.Name
-	routingtable, _ = RouteTableGen.GetID(name, 0)
+	routingtable = RouteTableGen.GetID(name)
 	log.Printf("LGM assigned id %+v for vrf name %s\n", routingtable, vrf.Name)
 	isbusy, err := routingtableBusy(routingtable)
 	if err != nil {
--- a/pkg/LinuxGeneralModule/lgm.go
+++ b/pkg/LinuxGeneralModule/lgm.go
@@ -31,12 +31,6 @@ import (
 // ModulelgmHandler enmpty interface
 type ModulelgmHandler struct{}
 
-// RoutingTableMax max value of routing table
-const RoutingTableMax = 4000
-
-// RoutingTableMin min value of routing table
-const RoutingTableMin = 1000
-
 // lgmComp string constant
 const lgmComp string = "lgm"
 
@@ -46,10 +40,11 @@ const brStr string = "br-"
 // vxlanStr string constant
 const vxlanStr string = "vxlan-"
 
-// GenerateRouteTable range specification, note that min <= max
-func GenerateRouteTable() uint32 {
-	return uint32(rand.Intn(RoutingTableMax-RoutingTableMin+1) + RoutingTableMin) //nolint:gosec
-}
+// routingTableMax max value of routing table
+const routingTableMax = 4000
+
+// routingTableMin min value of routing table
+const routingTableMin = 1000
 
 // run runs the commands
 func run(cmd []string, flag bool) (string, int) {
@@ -359,6 +354,9 @@ var ctx context.Context
 // nlink variable wrapper
 var nlink utils.Netlink
 
+// RouteTableGen table id generate variable
+var RouteTableGen utils.IDPool
+
 // Initialize initializes the config, logger and subscribers
 func Initialize() {
 	eb := eventbus.EBus
@@ -372,7 +370,8 @@ func Initialize() {
 	brTenant = "br-tenant"
 	ipMtu = config.GlobalConfig.LinuxFrr.IPMtu
 	ctx = context.Background()
-	nlink = utils.NewNetlinkWrapperWithArgs(config.GlobalConfig.Tracer)
+	RouteTableGen = utils.IDPoolInit("RTtable", routingTableMin, routingTableMax)
+	nlink = utils.NewNetlinkWrapperWithArgs(false)
 	// Set up the static configuration parts
 	_, err := nlink.LinkByName(ctx, brTenant)
 	if err != nil {
@@ -410,8 +409,8 @@ func setUpTenantBridge() {
 	}
 }
 
-// routingTableBusy checks if the route is in filterred list
-func routingTableBusy(table uint32) (bool, error) {
+// routingtableBusy checks if the route is in filterred list
+func routingtableBusy(table uint32) (bool, error) {
 	routeList, err := nlink.RouteListFiltered(ctx, netlink.FAMILY_V4, &netlink.Route{Table: int(table)}, netlink.RT_FILTER_TABLE)
 	if err != nil {
 		return false, err
@@ -463,6 +462,7 @@ func setUpBridge(lb *infradb.LogicalBridge) (string, bool) {
 //nolint:funlen,gocognit
 func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	IPMtu := fmt.Sprintf("%+v", ipMtu)
+	var addKey int
 	if path.Base(vrf.Name) == "GRD" {
 		vrf.Metadata.RoutingTable = make([]*uint32, 2)
 		vrf.Metadata.RoutingTable[0] = new(uint32)
@@ -473,19 +473,23 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	}
 	vrf.Metadata.RoutingTable = make([]*uint32, 1)
 	vrf.Metadata.RoutingTable[0] = new(uint32)
-	var routingTable uint32
+	var routingtable uint32
+	name := vrf.Name
 	for {
-		routingTable = GenerateRouteTable()
-		isBusy, err := routingTableBusy(routingTable)
+		routingtable, _ = RouteTableGen.GetID(name, 0)
+		log.Printf("LGM assigned id %+v for vrf name %s\n", routingtable, vrf.Name)
+		isbusy, err := routingtableBusy(routingtable)
 		if err != nil {
-			log.Printf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingTable, err)
-			return fmt.Sprintf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingTable, err), false
+			log.Printf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingtable, err)
+			return fmt.Sprintf("LGM : Error occurred when checking if routing table %d is busy: %+v\n", routingtable, err), false
 		}
-		if !isBusy {
-			log.Printf("LGM: Routing Table %d is not busy\n", routingTable)
+		if !isbusy {
+			log.Printf("LGM: Routing Table %d is not busy\n", routingtable)
 			break
 		}
-		log.Printf("LGM: Routing Table %d is busy\n", routingTable)
+		log.Printf("LGM: Routing Table %d is busy\n", routingtable)
+		addKey++
+		name = fmt.Sprintf("%s%d", name, addKey)
 	}
 	var vtip string
 	if !reflect.ValueOf(vrf.Spec.VtepIP).IsZero() {
@@ -498,19 +502,19 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 			return fmt.Sprintf(" LGM: VTEP IP not found: %+v\n", vrf.Spec.VtepIP), false
 		}
 	}
-	log.Printf("setUpVrf: %s %d\n", vtip, routingTable)
+	log.Printf("setUpVrf: %s %d\n", vtip, routingtable)
 	// Create the vrf interface for the specified routing table and add loopback address
 
 	linkAdderr := nlink.LinkAdd(ctx, &netlink.Vrf{
 		LinkAttrs: netlink.LinkAttrs{Name: path.Base(vrf.Name)},
-		Table:     routingTable,
+		Table:     routingtable,
 	})
 	if linkAdderr != nil {
-		log.Printf("LGM: Error in Adding vrf link table %d\n", routingTable)
-		return fmt.Sprintf("LGM: Error in Adding vrf link table %d\n", routingTable), false
+		log.Printf("LGM: Error in Adding vrf link table %d\n", routingtable)
+		return fmt.Sprintf("LGM: Error in Adding vrf link table %d\n", routingtable), false
 	}
 
-	log.Printf("LGM: vrf link %s Added with table id %d\n", vrf.Name, routingTable)
+	log.Printf("LGM: vrf link %s Added with table id %d\n", vrf.Name, routingtable)
 
 	link, linkErr := nlink.LinkByName(ctx, path.Base(vrf.Name))
 	if linkErr != nil {
@@ -545,7 +549,7 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 
 	Src1 := net.IPv4(0, 0, 0, 0)
 	route := netlink.Route{
-		Table:    int(routingTable),
+		Table:    int(routingtable),
 		Type:     unix.RTN_THROW,
 		Protocol: 255,
 		Priority: 9999,
@@ -557,7 +561,7 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 		return fmt.Sprintf("LGM : Failed in adding Route throw default %+v\n", routeaddErr), false
 	}
 
-	log.Printf("LGM : Added route throw default table %d proto opi_evpn_br metric 9999\n", routingTable)
+	log.Printf("LGM : Added route throw default table %d proto opi_evpn_br metric 9999\n", routingtable)
 	// Disable reverse-path filtering to accept ingress traffic punted by the pipeline
 	// disable_rp_filter("rep-"+vrf.Name)
 	// Configuration specific for VRFs associated with L3 EVPN
@@ -647,8 +651,8 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 			return fmt.Sprintf("LGM : Unable to set link %s UP \n", vrf.Name), false
 		}
 	}
-	details := fmt.Sprintf("{\"routingTable\":\"%d\"}", routingTable)
-	*vrf.Metadata.RoutingTable[0] = routingTable
+	details := fmt.Sprintf("{\"routingtable\":\"%d\"}", routingtable)
+	*vrf.Metadata.RoutingTable[0] = routingtable
 	return details, true
 }
 
@@ -784,7 +788,7 @@ func tearDownVrf(vrf *infradb.Vrf) (string, bool) {
 	if path.Base(vrf.Name) == "GRD" {
 		return "", true
 	}
-	routingTable := *vrf.Metadata.RoutingTable[0]
+	routingtable := *vrf.Metadata.RoutingTable[0]
 	// Delete the Linux networking artefacts in reverse order
 	if !reflect.ValueOf(vrf.Spec.Vni).IsZero() {
 		linkVxlan, linkErr := nlink.LinkByName(ctx, vxlanStr+path.Base(vrf.Name))
@@ -811,7 +815,7 @@ func tearDownVrf(vrf *infradb.Vrf) (string, bool) {
 		}
 		log.Printf("LGM : Delete br-%s\n", vrf.Name)
 	}
-	routeTable := fmt.Sprintf("%+v", routingTable)
+	routeTable := fmt.Sprintf("%+v", routingtable)
 	flusherr := nlink.RouteFlushTable(ctx, routeTable)
 	if flusherr != nil {
 		log.Printf("LGM: Error in flush table  %+v\n", routeTable)
--- a/pkg/utils/idpool.go
+++ b/pkg/utils/idpool.go
@@ -37,24 +37,24 @@ type IDPool struct {
 }
 
 // IDPoolInit initialize mod ptr pool
-func IDPoolInit(name string, min uint32, max uint32) IDPool {
+func IDPoolInit(name string, min uint32, max uint32) (IDPool, bool) {
+	if max < min {
+		log.Printf("IDPool: Failed to Init pool for %s\n", name)
+		return IDPool{}, false
+	}
 	var pool IDPool
 	pool.name = name
 	var index int
 	pool.unusedIDs = make([]uint32, (max-min)+1)
-	if min > 0 {
-		for value := max; value >= min; value-- {
-			pool.unusedIDs[index] = value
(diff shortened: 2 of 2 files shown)


pkg/LinuxGeneralModule/lgm.go
```diff
@@ -465,8 +467,12 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	vrf.Metadata.RoutingTable = make([]*uint32, 1)
 	vrf.Metadata.RoutingTable[0] = new(uint32)
 	var routingTable uint32
+	Name := vrf.Name
```
artek-koltun, line 470: please align with Go style guide:
https://google.github.io/styleguide/go/decisions.html

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -465,8 +467,12 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
...
+	addKey = 0
 	for {
-		routingTable = GenerateRouteTable()
+		// var key interface{}
```
artek-koltun, line 473: if code is not sued, it can be removed

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -465,8 +467,12 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
...
+		routingTable, _ = RouteTableGen.GetID(Name, 0)
```
artek-koltun, line 474: why the second returned valued is ignored?
why the second arg is always 0?

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -477,6 +483,8 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 			break
 		}
 		log.Printf("LGM: Routing Table %d is busy\n", routingTable)
+		addKey++
+		Name = fmt.Sprintf("%s%d", Name, addKey)
```
artek-koltun, line 487: 1. style
2. If we use string concatenation to create names, i'd prefer it to be extracted into a separate function with clear name, what kind of id we cretate here

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+// Copyright (C) 2023 Nordix Foundation.
+
+// Package linuxgeneralmodule is the main package of the application
+
+package utils
+
+import (
+	"log"
+	"reflect"
+	// "github.com/opiproject/opi-evpn-bridge/pkg/vendor_plugins/intel-e2000/p4runtime/p4translation"
```
artek-koltun, line 12: if code is not sued, it can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
+// SPDX-License-Identifier: Apache-2.0
```
artek-koltun, line 1: name of the file must be in lower case

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+#   released the ID with its reference. When a reference is specified in get_id()
+#   and release_id() the IDPool returns the current number of reference for the
+#   ID so that a caller knows when an ID was newly assigned (ref_count 1) or
+#   finally released (ref_count 0).
+# ---------------------------------------------------------------------------------
+*/
+type IDPool struct {
+	// self._lock = threading.Lock()
+	name         string                 // Name of pool
+	_unusedIDs   []uint32               // Yet unused IDs in pool Available ids
```
artek-koltun, line 35: style for member naming

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+#   to support multiple independent clients.
+#   The pool will only release the ID for the key, when the last client has the
+#   released the ID with its reference. When a reference is specified in get_id()
+#   and release_id() the IDPool returns the current number of reference for the
+#   ID so that a caller knows when an ID was newly assigned (ref_count 1) or
+#   finally released (ref_count 0).
+# ---------------------------------------------------------------------------------
+*/
+type IDPool struct {
+	// self._lock = threading.Lock()
```
artek-koltun, line 33: if code is not sued, it can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	_unusedIDs   []uint32               // Yet unused IDs in pool Available ids
+	_idsInUse    map[interface{}]uint32 // Mapping key: id for currently assigned ids
+	_idsForReuse map[interface{}]uint32 // Mapping key: id for previously assigned ids
+	_refs        map[uint32][]interface{}
+	_size        int // Size of the pool
+}
+
+// IDPoolInit initialize mod ptr pool
+func IDPoolInit(name string, min uint32, max uint32) IDPool {
+	var id IDPool
```
artek-koltun, line 44: it is rather pool than id

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	id.name = name
```
artek-koltun, line 45: is name used anywhere?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	if !reflect.ValueOf(id).IsZero() {
+		ip._idsInUse[key] = id
+	}
+	return id
+}
+
+// GetID get the mod ptr id from pool
+func (ip *IDPool) GetID(key interface{}, ref interface{}) (uint32, uint32) {
+	id := ip._idsInUse[key]
+	if reflect.ValueOf(id).IsZero() {
```
artek-koltun, line 92: why not more conventional
```go
_, ok := ip._idsInUse[key]
if !ok {
}
```

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		// Assign a free id for the key
```
artek-koltun, line 93: comment is not needed, the function name states the same

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	id._idsInUse = make(map[interface{}]uint32)
+	id._idsForReuse = make(map[interface{}]uint32)
+	id._refs = make(map[uint32][]interface{})
+	return id
+}
+
+func (ip *IDPool) _assignID(key interface{}) uint32 {
+	// Check if there was an id assigned for that key earlier
+	id := ip._idsForReuse[key]
+	if !reflect.ValueOf(id).IsZero() {
```
artek-koltun, line 60: `id != 0` ?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		// Re-use the old id
+		delete(ip._idsForReuse, key)
+	} else {
+		if len(ip._unusedIDs) != 0 {
+			// Pick an unused id
+			id = ip._unusedIDs[0]
+			ip._unusedIDs = append(ip._unusedIDs[:0], ip._unusedIDs[1:]...)
```
artek-koltun, line 67: why not pick from the end? Removing from element is much slower(yes, it is control path, but anyway, picking from the end is much more effective)

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	"log"
+	"reflect"
+	// "github.com/opiproject/opi-evpn-bridge/pkg/vendor_plugins/intel-e2000/p4runtime/p4translation"
+)
+
+// IDPool structure
+/*  Helper class for uniquely assigning IDs from a specified integer set (e.g. a
+#   range) to keys. IDs are assigned (or read) with get_id(key) and returned back
+#   into the pool with release_id(key). The IDPool remembers a once-assigned ID
+#   for keys so that the same ID is assigned for a key. Only when the pool runs
```
artek-koltun, line 19: why do we need such convoluted logic? Why we cannot pick any free id?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+			ip._unusedIDs = append(ip._unusedIDs[:0], ip._unusedIDs[1:]...)
+		} else {
+			if len(ip._idsForReuse) != 0 {
+				// Pick one of the ids earlier used for another key
+				for oldKey := range ip._idsForReuse {
+					delete(ip._idsForReuse, oldKey)
+					break
+				}
+			} else {
+				// No id left
```
artek-koltun, line 76: comment not needed, it is clear from the code

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+				log.Printf("IDPool: Failed to allocate id for %+v. No free ids in pool.", key)
+				return 0
+			}
+		}
+	}
+	// Store the assigned id, if any
+	if !reflect.ValueOf(id).IsZero() {
```
artek-koltun, line 83: `id != 0`?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+			if len(ip._idsForReuse) != 0 {
+				// Pick one of the ids earlier used for another key
+				for oldKey := range ip._idsForReuse {
+					delete(ip._idsForReuse, oldKey)
+					break
+				}
+			} else {
+				// No id left
+				log.Printf("IDPool: Failed to allocate id for %+v. No free ids in pool.", key)
+				return 0
```
artek-koltun, line 78: what if it was requested a pool with min=0 and max=42?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	if reflect.ValueOf(id).IsZero() {
+		// Assign a free id for the key
+		id = ip._assignID(key)
+		if id == 0 {
+			return 0, 0
+		}
+	}
+	if !reflect.ValueOf(ref).IsZero() {
+		log.Printf("IDPool: GetID  Assigning key : %+v , id  %+v for ref %v", id, key, ref)
+		// refSet := ip._refs[id]
```
artek-koltun, line 101: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		if reflect.ValueOf(ip._refs[id]).IsZero() {
```
artek-koltun, line 102: `!= nil`

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		ip._refs[id] = append(ip._refs[id], ref)
+		return id, uint32(len(ip._refs[id]))
+	}
+	log.Printf("IDPool: GetID Assigning id %v for key %v and ref %v", id, key, ref)
+	return id, uint32(0)
+}
+
+func deleteRef(refSet []interface{}, ref interface{}) []interface{} {
+	// size := len(refSet)
```
artek-koltun, line 113: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+			i = uint32(index)
+			break
+		}
+	}
+	return append(refSet[:i], refSet[i+1:]...)
+}
+
+// ReleaseID get the reference id
+func (ip *IDPool) ReleaseID(key interface{}, ref interface{}) (uint32, uint32) {
+	// with self._lock:
```
artek-koltun, line 126: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	log.Printf("IDPool:ReleaseID  Releasing id for key %v", key)
+	id := ip._idsInUse[key]
+	if reflect.ValueOf(ref).IsZero() {
```
artek-koltun, line 129: == nil

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		log.Printf("No id to release for key %v", key)
+		return 0, 0
+	}
+	refSet := ip._refs[id]
+	if !reflect.ValueOf(refSet).IsZero() && !reflect.ValueOf(ref).IsZero() {
```
artek-koltun, line 134: != nil

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+func deleteRef(refSet []interface{}, ref interface{}) []interface{} {
+	// size := len(refSet)
+	var i uint32
+	for index, value := range refSet {
+		if value == ref {
+			i = uint32(index)
+			break
+		}
+	}
+	return append(refSet[:i], refSet[i+1:]...)
```
artek-koltun, line 121: if you need to perform such operations, why not use `map[key]struct{}`?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	if reflect.ValueOf(ref).IsZero() {
+		log.Printf("No id to release for key %v", key)
+		return 0, 0
+	}
+	refSet := ip._refs[id]
+	if !reflect.ValueOf(refSet).IsZero() && !reflect.ValueOf(ref).IsZero() {
+		// Remove the specified reference from the id
+		refSet = deleteRef(refSet, ref)
+	}
+	if !reflect.ValueOf(refSet).IsZero() {
```
artek-koltun, line 138: != nil

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		// No (remaining) references, release id
```
artek-koltun, line 139: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		log.Printf("IDPool:ReleaseID Id %v has been released", id)
+		delete(ip._idsInUse, key)
+		if !reflect.ValueOf(refSet).IsZero() {
```
artek-koltun, line 142: !=nil

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+			delete(ip._refs, id)
+		}
+		// Store released id for future reassignment
+		ip._idsForReuse[key] = id
+	} else {
+		log.Printf("IDPool:ReleaseID Keep id:%+v remaining references %+v", id, len(refSet))
+	}
+	if !reflect.ValueOf(ref).IsZero() {
```
artek-koltun, line 150: != nil

Is this condition needed at all?

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+	// with self._lock:
+	log.Printf("IDPool:ReleaseID  Releasing id for key %v", key)
+	id := ip._idsInUse[key]
+	if reflect.ValueOf(ref).IsZero() {
+		log.Printf("No id to release for key %v", key)
+		return 0, 0
+	}
+	refSet := ip._refs[id]
+	if !reflect.ValueOf(refSet).IsZero() && !reflect.ValueOf(ref).IsZero() {
+		// Remove the specified reference from the id
```
artek-koltun, line 135: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		refSet = deleteRef(refSet, ref)
+	}
+	if !reflect.ValueOf(refSet).IsZero() {
+		// No (remaining) references, release id
+		log.Printf("IDPool:ReleaseID Id %v has been released", id)
+		delete(ip._idsInUse, key)
+		if !reflect.ValueOf(refSet).IsZero() {
+			delete(ip._refs, id)
+		}
+		// Store released id for future reassignment
```
artek-koltun, line 145: can be removed

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+		if value == ref {
+			i = uint32(index)
+			break
+		}
+	}
+	return append(refSet[:i], refSet[i+1:]...)
+}
+
+// ReleaseID get the reference id
+func (ip *IDPool) ReleaseID(key interface{}, ref interface{}) (uint32, uint32) {
```
artek-koltun, line 125: I do not see that ReleaseID is used anywhere

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	"log"
+	"reflect"
+	// "github.com/opiproject/opi-evpn-bridge/pkg/vendor_plugins/intel-e2000/p4runtime/p4translation"
+)
+
+// IDPool structure
+/*  Helper class for uniquely assigning IDs from a specified integer set (e.g. a
+#   range) to keys. IDs are assigned (or read) with get_id(key) and returned back
```
artek-koltun, line 17: there is no get_id(key) function

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+#   into the pool with release_id(key). The IDPool remembers a once-assigned ID
```
artek-koltun, line 18: there is no `release_id(key)` function

pkg/utils/Idpool.go
```diff
@@ -0,0 +1,154 @@
...
+#   finally released (ref_count 0).
+# ---------------------------------------------------------------------------------
+*/
+type IDPool struct {
+	// self._lock = threading.Lock()
+	name         string                 // Name of pool
+	_unusedIDs   []uint32               // Yet unused IDs in pool Available ids
+	_idsInUse    map[interface{}]uint32 // Mapping key: id for currently assigned ids
+	_idsForReuse map[interface{}]uint32 // Mapping key: id for previously assigned ids
+	_refs        map[uint32][]interface{}
```
artek-koltun, line 38: refs sounds like a counter tracking how many times we requested the same id, no?

pkg/LinuxGeneralModule/lgm.go
venkyvsp: second returned value can be ignored if the second argument is 0. 
second argument is for the reference to the same key.   actually in the routing table id scenario, there is not need for reference hence ignored.
In vendor code, we used reference for ECMP id case

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -46,9 +40,12 @@ const brStr string = "br-"
 // vxlanStr string constant
 const vxlanStr string = "vxlan-"
 
-// GenerateRouteTable range specification, note that min <= max
-func GenerateRouteTable() uint32 {
-	return uint32(rand.Intn(RoutingTableMax-RoutingTableMin+1) + RoutingTableMin) //nolint:gosec
+// RoutingTableRange ModPointer structure of  mod ptr definitions
+var RoutingTableRange = struct {
```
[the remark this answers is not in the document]
venkyvsp: Done

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -363,7 +363,8 @@ func Initialize() {
 	brTenant = "br-tenant"
 	ipMtu = config.GlobalConfig.LinuxFrr.IPMtu
 	ctx = context.Background()
-	nlink = utils.NewNetlinkWrapperWithArgs(config.GlobalConfig.Tracer)
+	RouteTableGen = utils.IDPoolInit("RTtable", RoutingTableRange.RoutingTableMin, RoutingTableRange.RoutingTableMax)
```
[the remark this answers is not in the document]
venkyvsp: Done

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -454,6 +455,7 @@ func setUpBridge(lb *infradb.LogicalBridge) bool {
 //nolint:funlen,gocognit
 func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 	IPMtu := fmt.Sprintf("%+v", ipMtu)
+	var addKey int
```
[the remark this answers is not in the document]
venkyvsp: Modified

pkg/LinuxGeneralModule/lgm.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Yes it is used in the lgm.go and vendor plugin code.

pkg/utils/Idpool.go
venkyvsp: its used in vendor plugin code dcgw.go for releasing the ecmp and trie id's

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: correct, its reference to the same id

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: It will create a pool of 43's id from the 0 to 42 it includes 42 as well.

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: It is for deleting a ref element from the refset(array) pointing to an id index.
we feel array is preferable instead of struct

pkg/utils/Idpool.go
venkyvsp: The crucial feature of this ID allocator is that a unique DB object (as defined by its key fields) always gets the same ID assigned, no matter how often we request an ID. For the netlink watcher, we request an ID for the same objects every scanning interval. If it got completely independent IDs every time, it would be impossible to compare two subsequent Netlink DB shapshots and calculate the delta for generating add/modify/delete notifications.

pkg/LinuxGeneralModule/lgm.go
artek-koltun: if we use concatenation to form names, I'd prefer to get it extracted into a function which would have explanatory name, what kind of name we are constructing here

pkg/utils/Idpool.go
artek-koltun: Member names cannot start with `_`

pkg/utils/Idpool.go
artek-koltun: in lgm.go we create pool with a given name, but name is not exported and as a member, it is not used in idpool.go
apparently it can be removed

pkg/utils/Idpool.go
artek-koltun: it is reworked to 
```go
        id := ip._idsinuse[key]
	if id == 0 {
		id = ip._assignID(key)
		if id == 0 {
			return 0, 0
		}
	}
```

but I do not see that in init we check that min should be greater than 0

pkg/utils/Idpool.go
artek-koltun: https://github.com/opiproject/opi-evpn-bridge/pull/400#discussion_r1810760434

pkg/utils/Idpool.go
artek-koltun: still do not understand why we cannot use `map[key]struct{}` instead of `refSet []interface{}`

pkg/utils/Idpool.go
artek-koltun: then why do we need a slice of slices then? Can we have a slice of counters? `map[uint32]uint32`?

pkg/utils/idpool.go
```diff
@@ -0,0 +1,144 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022-2023 Intel Corporation, or its subsidiaries.
+// Copyright (C) 2023 Nordix Foundation.
+
+// Package linuxgeneralmodule is the main package of the application
```
artek-koltun, line 5: ?

pkg/utils/Idpool.go
artek-koltun: you are still picking the first element from the slice, which is suboptimal? If you reverse your unused id in init and will pick last element (which is more effective)?

pkg/utils/Idpool.go
venkyvsp: Name is used for debug purpose for printing Pool status info.
Added new function for printing Pool info with name.  Thanks

pkg/utils/Idpool.go
venkyvsp: actually we need golang type equivalent of 'list' type in python.  
we have replaced with "ref map[uint32]map[interface{}]bool"
here the scenario is for same key we can have different references of interface type as reference can be of any type like int, string, alphanumeric etc. that's why we cannot use counter here.

pkg/utils/Idpool.go
venkyvsp: actually we need golang type equivalent of 'list' type in python.
we have replaced with "ref map[uint32]map[interface{}]bool"
here the scenario is for same key we can have different references of interface type as reference can be of any type like int, string, alphanumeric etc. that's why we cannot use counter here.

pkg/utils/Idpool.go
venkyvsp: done

pkg/LinuxGeneralModule/lgm.go
venkyvsp: removed the concatenation code. if the table id is busy we return error.

pkg/LinuxGeneralModule/lgm.go
venkyvsp: Done

pkg/utils/Idpool.go
venkyvsp: Used "ok"  and also checking the minimum value in  the init function

pkg/utils/Idpool.go
venkyvsp: checking the minimum value in the init function

pkg/utils/idpool.go
venkyvsp: done

pkg/LinuxGeneralModule/lgm.go
```diff
@@ -363,7 +361,12 @@ func Initialize() {
 	brTenant = "br-tenant"
 	ipMtu = config.GlobalConfig.LinuxFrr.IPMtu
 	ctx = context.Background()
-	nlink = utils.NewNetlinkWrapperWithArgs(config.GlobalConfig.Tracer)
+	RouteTableGen = utils.IDPoolInit("RTtable", routingTableMin, routingTableMax)
```
artek-koltun, line 364: `IDPoolInit` can just return `error` if failed to initialize

pkg/LinuxGeneralModule/lgm.go
artek-koltun: then maybe we could have 2 functions `GetID` and `GetIDWithRef`?

pkg/utils/Idpool.go
artek-koltun: if you declare only one variable to get the result of map lookup, then it is not `ok`. It is value with the given key or zero for integers otherwise. Your current form with ok looks suspicious. I think you need to use `_, ok := ip._idsInUse[key]` to check presense

I see you added this into init
```go
if min > 0 {
		for value := max; value >= min; value-- {
			pool.unusedIDs[index] = value
			index++
		}
	} else {
		return IDPool{}
	}
```
What would happen if I call it like this?
```go
pool := IDPoolInit("my_pool", 1, 0)
```

pkg/utils/Idpool.go
artek-koltun: https://github.com/opiproject/opi-evpn-bridge/pull/400/#discussion_r1876231277

pkg/utils/Idpool.go
artek-koltun: https://github.com/opiproject/opi-evpn-bridge/pull/400/#discussion_r1876231277

pkg/LinuxGeneralModule/lgm.go
venkyvsp: added bool return for the initialization status

pkg/utils/Idpool.go
venkyvsp: modified for checking the existence of value in the map for the key with ok.  and 
Also addressed the scenario of min is greater that max value. 
for if min > 0 , linter is throwing error that min is uint32 never be negative value.

pkg/utils/Idpool.go
venkyvsp: done modified for checking the existence of value in the map for the key with ok.

pkg/utils/Idpool.go
venkyvsp: modified for checking the existence of value in the map for the key with ok.

pkg/utils/idpool.go
```diff
@@ -0,0 +1,172 @@
...
+		log.Printf("IDPool: GetID Assigning id %v for key %v ", id, key)
+	}
+	return id
+}
+
+// GetIDWithRef get the mod ptr id from pool with Referenbce
+func (ip *IDPool) GetIDWithRef(key interface{}, ref interface{}) (uint32, uint32) {
+	var ok bool
+	var id uint32
+	if id, ok = ip.idsInUse[key]; !ok {
```
artek-koltun, line 115: why not reuse GetID?

pkg/utils/Idpool.go
artek-koltun: actually, I don't like it. map of maps sounds too complicated. Pls try to come up with smth simpler

pkg/utils/idpool.go
venkyvsp: Done

artek-koltun (approved):

PR #401, opened 2024-09-25
fix(evpn-bridge): fix for frr to handle vrf setup and teardown

venkyvsp:
fix: frr handle the duplicate vni and deleting of non-existence vrf and svi.

### DIFF
--- a/pkg/LinuxCIModule/lci.go
+++ b/pkg/LinuxCIModule/lci.go
@@ -7,6 +7,7 @@ package linuxcimodule
 
 import (
 	"context"
+	"fmt"
 
 	// "io/ioutil"
 	"log"
@@ -48,6 +49,7 @@ func handlebp(objectData *eventbus.ObjectData) {
 		log.Printf("LCI : GetBP error: %s\n", err)
 		comp.Name = lciComp
 		comp.CompStatus = common.ComponentStatusError
+		comp.Details = fmt.Sprintf("LCI : GetBP error: %s\n", err)
 		if comp.Timer == 0 {
 			comp.Timer = 2 * time.Second
 		} else {
@@ -63,6 +65,7 @@ func handlebp(objectData *eventbus.ObjectData) {
 		log.Printf("LVM: Mismatch in resoruce version %+v\n and bp resource version %+v\n", objectData.ResourceVersion, BP.ResourceVersion)
 		comp.Name = lciComp
 		comp.CompStatus = common.ComponentStatusError
+		comp.Details = fmt.Sprintf("LVM: Mismatch in resoruce version %+v\n and bp resource version %+v\n", objectData.ResourceVersion, BP.ResourceVersion)
 		if comp.Timer == 0 {
 			comp.Timer = 2 * time.Second
 		} else {
@@ -82,10 +85,10 @@ func handlebp(objectData *eventbus.ObjectData) {
 		}
 	}
 	if BP.Status.BPOperStatus != infradb.BridgePortOperStatusToBeDeleted {
-		status := setUpBp(BP)
+		details, status := setUpBp(BP)
 		comp.Name = lciComp
+		comp.Details = details
 		if status {
-			comp.Details = ""
 			comp.CompStatus = common.ComponentStatusSuccess
 			comp.Timer = 0
 		} else {
@@ -102,8 +105,9 @@ func handlebp(objectData *eventbus.ObjectData) {
 			log.Printf("error in updating bp status: %s\n", err)
 		}
 	} else {
-		status := tearDownBp(BP)
+		details, status := tearDownBp(BP)
 		comp.Name = lciComp
+		comp.Details = details
 		if status {
 			comp.CompStatus = common.ComponentStatusSuccess
 			comp.Timer = 0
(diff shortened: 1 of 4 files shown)


pkg/frr/frr.go
```diff
@@ -514,12 +510,17 @@ func tearDownVrf(vrf *infradb.Vrf) bool {
 		log.Printf("FRR Deleted event")
 		delCmd1 := fmt.Sprintf("no router bgp %+v vrf %s", localas, path.Base(vrf.Name))
 		delCmd2 := fmt.Sprintf("no vrf %s", path.Base(vrf.Name))
-		_, err = Frr.FrrBgpCmd(ctx, fmt.Sprintf("configure terminal\n %s\n exit\n", delCmd1))
-		if err != nil {
+		data, err = Frr.FrrBgpCmd(ctx, fmt.Sprintf("configure terminal\n %s\n exit\n", delCmd1))
+		if strings.Contains(data, "Can't find BGP instance") { // Trying to delete non exist VRF return true
+			return true
+		}
+		if err != nil || checkFrrResult(data, false) {
```
artek-koltun, line 517: if we need to check frr for each frr cmd, why not move that check into FrrBgpCmd/FrrZebraCmd and other frr functions?

as well as `strings.Contains(`?

pkg/frr/frr.go
```diff
@@ -514,12 +510,17 @@ func tearDownVrf(vrf *infradb.Vrf) bool {
...
+			log.Printf("FRR: Error  %s\n", data)
 			return false
 		}
-		_, err = Frr.FrrZebraCmd(ctx, fmt.Sprintf("configure terminal\n %s\n exit\n", delCmd2))
-		if err != nil {
+		data, err = Frr.FrrZebraCmd(ctx, fmt.Sprintf("configure terminal\n %s\n exit\n", delCmd2))
+		if err != nil || checkFrrResult(data, false) {
+			log.Printf("FRR: Error  %s\n", data)
```
artek-koltun, line 523: Does data contain error description?

pkg/frr/frr.go
```diff
@@ -437,7 +430,7 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
 
 // checkFrrResult checks the vrf result
 func checkFrrResult(cp string, show bool) bool {
-	return ((show && reflect.ValueOf(cp).IsZero()) || strings.Contains(cp, "warning") || strings.Contains(cp, "unknown") || strings.Contains(cp, "Unknown") || strings.Contains(cp, "Warning") || strings.Contains(cp, "Ambiguous") || strings.Contains(cp, "specified does not exist"))
+	return ((show && reflect.ValueOf(cp).IsZero()) || strings.Contains(cp, "warning") || strings.Contains(cp, "unknown") || strings.Contains(cp, "Unknown") || strings.Contains(cp, "Warning") || strings.Contains(cp, "Ambiguous") || strings.Contains(cp, "specified does not exist") || strings.Contains(cp, "Error"))
```
artek-koltun, line 433: reflect.ValueOf(cp).IsZero()) -> cp == ""

what if cp contains `waRning`? Will you add a new comparison then? Will it be better to lower case the string and only then check what itcontains?

pkg/frr/frr.go
```diff
@@ -379,24 +372,24 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
...
+		if err != nil || checkFrrResult(cp, true) {
+			log.Printf("FRR Error-show bgp l2vpn evpn vni %v cp %v", err, cp)
 		}
 		hname, _ := os.Hostname()
 		L2vpnCmd := strings.Split(cp, "json")
 		L2vpnCmd = strings.Split(L2vpnCmd[1], hname)
 		cp = L2vpnCmd[0]
-		// fmt.Printf("FRR_L2vpn[0]: %s\n",cp)
-		if len(cp) != 7 {
+		if len(cp) != 7 { // Checking CMD o/p
```
artek-koltun, line 392: magic number. Rely on string length is very unreliable

pkg/frr/frr.go
venkyvsp: Yes data contains the complete FRR command output.

pkg/frr/frr.go
venkyvsp: Addressed this and changed to string compare instead of string length

pkg/frr/frr.go
atulpatel261194: Done

pkg/frr/frr.go
venkyvsp: Done

pkg/frr/frr.go
artek-koltun: is response should be in json format? Should we better try to Unmarshall it then using`"\r\n{}\r\n{\r\n}\r\n"`? If not, maybe we could check if it contains alphanumerical values instead?

pkg/frr/frr.go
artek-koltun: would it be better in case of an error, in FrrBgpCmd, just add data content into err and return empty first argument?

pkg/frr/frr.go
venkyvsp: Changed all local vars names in frr.go and replaced all reflect.valueof IsZero with simple comparison.

pkg/frr/frr.go
venkyvsp: Changed the comparing the string with alphanumerical.

pkg/frr/frr.go
atulpatel261194: Done

pkg/frr/frr.go
```diff
@@ -577,114 +559,110 @@ func setUpVrf(vrf *infradb.Vrf) (string, bool) {
...
+		return "", true
 	}
-	err = Frr.Save(ctx)
+	err = frr.Save(ctx)
 	if err != nil {
 		log.Printf("FRR(tearDownVrf): Failed to run save command: %v\n", err)
 	}
 	// Clean up FRR last
-	if !reflect.ValueOf(vrf.Spec.Vni).IsZero() {
+	if *vrf.Spec.Vni != 0 {
```
artek-koltun, line 643: what if vni is `nil`?

pkg/LinuxCIModule/lci.go
```diff
@@ -124,92 +128,92 @@ func handlebp(objectData *eventbus.ObjectData) {
 }
 
 // setUpBp sets up the bridge port
-func setUpBp(bp *infradb.BridgePort) bool {
+func setUpBp(bp *infradb.BridgePort) (string, bool) {
```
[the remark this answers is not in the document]
atulpatel261194: keeping the Status and details separate makes it more clear as Details are used for component specific details. which can con have error or other details i.e ip, mac etc. as well.

pkg/frr/frr.go
```diff
@@ -431,139 +461,91 @@ func DeInitialize() {
...
-	TableVersion  uint
-	RouterID      string
-	DefaultLocPrf uint
-	LocalAS       int
-	Routes        route
-}
-
 // setUpVrf sets up the vrf
+//
+//nolint:funlen,gocognit
```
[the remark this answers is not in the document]
atulpatel261194: fixed

pkg/frr/frr.go
venkyvsp: Done

pkg/frr/frr.go
artek-koltun: How the original output of FrrBgpCmd look like? I have an impression that you are trying to remove brackets and then add them back

venkyvsp: Below is the original output of FrrBgpCmd, we trim output for extracting data between the brackets for unmarshal the json o/p. 
	
**acc-1449# show bgp l2vpn evpn vni 3000 json
{
  "vni":3000,
  "type":"L3",
  "inKernel":"True",
  "rd":"10.120.3.1:2",
  "originatorIp":"10.3.3.3",
  "advertiseGatewayMacip":"n/a",
  "advertiseSviMacIp":"n/a",
  "advertisePip":"Enabled",
  "sysIP":"10.3.3.3",
  "sysMac":"00:e8:d4:70:8e:70",
  "rmac":"00:e8:d4:70:8e:70",
  "importRts":[
    "65003:3000"
  ],
  "exportRts":[
    "65003:3000"
  ]
}
acc-1449#**

artek-koltun: >  we trim output for extracting data between the brackets for unmarshal

Why? We can unmarshal with brackets, no?
https://godbolt.org/z/rxzWGr7q6

venkyvsp: Changes done

artek-koltun (approved):

PR #407, opened 2024-11-05
 Generate mock Apis

atulpatel261194:
Fix linter issues in mock Apis


jayPare:
Correct me if I'm wrong, but won't we have to redo all those steps next time we regenerate all the mocks using mockery? I'm asking since the modifications seem to be done on auto-generated file.

sandersms:
> Correct me if I'm wrong, but won't we have to redo all those steps next time we regenerate all the mocks using mockery? I'm asking since the modifications seem to be done on auto-generated files.

The makefile contains the "mock_generate" step which autogenerates the mocks that are in the PR based on the file that are located in pkg/utils.  These are the autogenerated files based on the changes in the pkg/utils versions.

sandersms (approved):
Updates to the mock files look good

PR #417, opened 2026-06-25
 Add IsGRD field to separate GRD VRF from user VRFs

### LINKED ISSUE #413: Separate the GRD Vrf from the rest of the VRFs by using a specific field
It is needed to add an extra field in the Vrf infradb struct (e.g. is_grd boolean) in order to be able to efficiently separate the GRD Vrf from the rest of the Vrfs.

Currently we do that  by just checking the name of the Vrf. This is inefficient as we need to leak name information to all the modules in our system and we are dependent on which name the user is using.

**Note**: We should not change the API for that as the user should not be able to create GRD VRFs. The system internally should only be responsible to create GRD Vrfs.  Potentially this can happen by specifing extra argument to the [NewVrfWithArgs](https://github.com/opiproject/opi-evpn-bridge/blob/main/pkg/infradb/vrf.go#L72) function.

vickysharma-prog:
## Summary
Fixes #413

Adds an explicit `IsGRD bool` field to the `Vrf` struct to efficiently identify the GRD VRF internally, replacing fragile name-based string comparisons scattered across multiple modules.

## Changes
- Add `IsGRD bool` field to `Vrf` struct in `pkg/infradb/vrf.go`
- Extend `NewVrfWithArgs()` with `isGRD bool` parameter — `NewVrf()` (protobuf/user path) remains completely unchanged
- Pass `isGRD=true` during GRD creation in `cmd/main.go`
- Replace all `path.Base(vrf.Name) == "GRD"` checks with `vrf.IsGRD`:
- `pkg/frr/frr.go` — setUpVrf, tearDownVrf
- `pkg/LinuxGeneralModule/lgm.go` — setUpVrf, tearDownVrf  
- `pkg/netlink/neighbor.go`
- `pkg/netlink/route.go`

## Testing
- `go build ./...` passes


vickysharma-prog:
@mardim91 @artek-koltun Hi! I've implemented the approach discussed in 
issue #413. This PR adds the `IsGRD` field to eliminate name-based GRD 
identification across all modules.

**Changes:**
- Add `IsGRD bool` field to `Vrf` struct
- Extend `NewVrfWithArgs()` with `isGRD` parameter (keeps `NewVrf()` API unchanged)
- Replace all `path.Base(vrf.Name) == "GRD"` string checks with `vrf.IsGRD`
- No changes to public API — users cannot create GRD VRFs

Ready for review! 🚀

